PR #2, opened 2015-09-08
Player is never initialized, and contentWindow is undefined

ruprict:
I had 2 issues using this lib:

1. In `MediaPlayer`, `this.state.Player` is never initialized for me `componentWillReceiveProps` is not called when I create the component. The component is part of a conditional render, so maybe that's the issue.
2. `iframe.contentWindow` is undefined for me. 

These two minor fixes allowed me to use the module. Any chance you could take a look and, if all's well, issue a new npm module?

### DIFF
--- a/src/ReactPlayer.js
+++ b/src/ReactPlayer.js
@@ -19,13 +19,13 @@ export default class MediaPlayer extends Component {
     return players.some(player => player.canPlay(url))
   }
   state = {
-    Player: null
+    Player:  players.find(Player => Player.canPlay(this.props.url))
   }
   componentWillReceiveProps (nextProps) {
     if (this.props.url !== nextProps.url) {
       this.setState({
         Player: players.find(Player => Player.canPlay(nextProps.url))
-      })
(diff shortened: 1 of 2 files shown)


CookPete:
@ruprict Thanks for your help! This is fixed in `0.0.7`. I hope you don't mind that I added some separate commits instead of merging this PR. `npm run lint` fails on your fork, so it just saved us going back and forth sorting that out.

`iframe.contentWindow` being undefined is quite worrying, as we then can't control the Vimeo player using external controls. What OS/browser did you see it happen on?

ruprict:
I am on Chrome 44.0.2403.157 (64-bit) and Mac OSX 10.10.5 (Yosemite). However, I think the `postMessage` function is called many times as the component is built, and it's only null up until the iframe is actually added to the DOM. My guess is while the iframe is in the virtual DOM, it's null. Once it's added to the real DOM, `contentWindow` is there.  It's a total guess, though.  Thoughts?

PR #4, opened 2015-10-18
Configuration

Fauntleroy:
#### Objective

Allow configuration of media players via `props`, specifically the `client_id` of the SoundCloud player.

-----

#### What I did

- Added the `soundcloudConfig`, `vimeoConfig`, and `youtubeConfig` props.
- Moved some configuration data into `defaultProps` accordingly.
- Ensured that every player would automatically set its volume on startup.
- Updated the debug page.

-----

#### How to verify

- Pull this branch
- Run `npm run start`
- Go to http://localhost:3000/
- Write custom configuration JSON and set it, such as:
```js
{
  "youtube": {
    "playerVars": {
      "autoplay": 0
    }
  }
}
```
- When initialized, the YouTube player should be modified accordingly. Note: This will not work on players that are already initialized.
- Modify the volume
- Switch to another player
- The player should be at the volume you set

-----

#### Screenshots

screen shot 2015-10-18 at 2 12 44 pm

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -38,6 +38,16 @@ export default class App extends Component {
       this.setState(state)
     }
   }
+  onConfigSubmit = () => {
+    let config
+    try {
+      config = JSON.parse(this.refs.config.value)
+    } catch (error) {
+      config = {}
+      console.error('Error setting config:', error)
+    }
+    this.setState(config)
+  }
   render () {
     return (
       <div>
@@ -48,6 +58,9 @@ export default class App extends Component {
           playing={this.state.playing}
           volume={this.state.volume}
           onProgress={this.onProgress}
+          soundcloudConfig={this.state.soundcloudConfig}
+          vimeoConfig={this.state.vimeoConfig}
+          youtubeConfig={this.state.youtubeConfig}
           onPlay={() => console.log('onPlay')}
           onPause={() => console.log('onPause')}
           onBuffer={() => console.log('onBuffer')}
@@ -61,20 +74,26 @@ export default class App extends Component {
         <button onClick={this.load.bind(this, 'http://clips.vorwaerts-gmbh.de/big_buck_bunny.mp4')}>MP4 video</button>
         <button onClick={this.load.bind(this, 'http://clips.vorwaerts-gmbh.de/big_buck_bunny.ogv')}>OGV video</button>
         <button onClick={this.load.bind(this, 'http://clips.vorwaerts-gmbh.de/big_buck_bunny.webm')}>WEBM video</button>
-        <input
+        <input ref='url' placeholder='url' />
+        <button onClick={() => { this.load(this.refs.url.value) }}>Load URL</button>
+        <hr />
+        seek: <input
           type='range' min={0} max={1} step='any'
           value={this.state.played}
           onMouseDown={this.onSeekMouseDown}
           onChange={this.onSeekChange}
           onMouseUp={this.onSeekMouseUp}
         />
-        <input
+        played: <progress max={1} value={this.state.played} />
+        loaded: <progress max={1} value={this.state.loaded} />
+        volume: <input
           type='range' min={0} max={1} step='any'
           value={this.state.volume}
           onChange={this.setVolume}
         />
-        played: <progress max={1} value={this.state.played} />
-        loaded: <progress max={1} value={this.state.loaded} />
+        <hr />
+        <textarea ref='config' placeholder='Config JSON' style={{width: '200px', height: '200px'}}></textarea>
+        <button onClick={this.onConfigSubmit}>Update Config</button>
       </div>
     )
   }
--- a/src/players/Vimeo.js
+++ b/src/players/Vimeo.js
@@ -67,6 +67,7 @@ export default class Vimeo extends Base {
       this.postMessage('addEventListener', 'pause')
       this.postMessage('addEventListener', 'finish')
     }
+    if (data.event === 'ready') this.onReady()
     if (data.event === 'playProgress') this.fractionPlayed = data.data.percent
     if (data.event === 'loadProgress') this.fractionLoaded = data.data.percent
     if (data.event === 'play') this.props.onPlay()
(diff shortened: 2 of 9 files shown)


Fauntleroy:
One important thing to note is that if you pass in a configuration object via props it will completely replace the default one. If preserving the other defaults while changing one thing is important to you, I can try to merge the configuration objects instead.

src/App.js
```diff
@@ -38,6 +38,15 @@ export default class App extends Component {
       this.setState(state)
     }
   }
+  onConfigSubmit = () => {
+    var config
```
CookPete: You should use `let` here I think

src/App.js
```diff
@@ -38,6 +38,15 @@ export default class App extends Component {
...
+    try {
+      config = JSON.parse(this.refs.config.value)
+    } catch (error) {
```
CookPete: No need for `error`?

src/players/YouTube.js
```diff
@@ -8,9 +8,19 @@ const SDK_URL = '//www.youtube.com/iframe_api'
...
+  autoplay: 1,
+  controls: 0,
+  showinfo: 0
+}
 
 export default class YouTube extends Base {
   static propTypes = propTypes
+  static defaultProps = {
+    youtube: {
+      playerVars: DEFAULT_PLAYER_VARS
```
CookPete: Maybe just inline the defaults here instead of using a `const`? To be consistent with the Vimeo player

CookPete:
Hi @Fauntleroy. Many thanks for your contribution!

I'm wondering if a prop for a config for each player is polluting the top level props too much, and maybe something like a `playerConfig` object to hold all of them would be better? Although then I suppose it could get awkward with default props and object merging for each player.

I also wonder if the deep structure is slightly overkill at this point, and whether we should just use `youtubePlayerVars`, `soundcloudClientId` and `vimeoFrameParams` instead.

CookPete:
Also, I've just pushed `0.1.0` with file support and some other minor bits which you will have to rebase on. Apologies!

Fauntleroy:
No problem, @CookPete! I'm more than happy to contribute to anything that'll rid me of JWPlayer (and the delicate `browserify-shim`s that come along with it).

Concerning the use of top-level props like `soundcloud`, `vimeo`, and `youtube`, I spent a while thinking about it last weekend and chose this approach for the following reasons:

- **Why props?** Using props to configure a react component seems like the best way to handle this (as opposed to a static like `Player.setConfig(...)`)
- **Why top-level?** These are pretty obscure configuration options that I don't expect most users of this component to utilize (aside from maybe the SoundCloud `client_id`, which you might want to emphasize). Since these props are optional, that means most users won't ever have to pollute their `render` methods with them.
- **Why not one `playerConfig` object?** Using separate props with keys related to each provider, in my opinion, is a cleaner and easier to use abstraction in practice.
- **Why deep objects?** These objects, while a bit unweildy, leave plenty of room for adding additional configuration options in the future. I'm going to resolve the issues with preserving `react-player` defaults by using `Object.assign` (or similar) to merge those defaults properly when the `SoundCloud`, `Vimeo`, or `YouTube` player is initialized.

src/App.js
Fauntleroy: Not that I believe this makes any sense, but `standard` throws on this line if you do not include `error`, even though it is ultimately unused:

```
standard: Use JavaScript Standard Style (https://github.com/feross/standard) 
  /Users/Tim/Repos/react-player/src/App.js:45:14: Parsing error: Unexpected token
```

Fauntleroy:
I responded to the code review comments, updated the way default `playerVars` and `iframeParams` are handled, and fixed the initial volume issue with the new `FilePlayer`.

src/players/Vimeo.js
```diff
@@ -6,9 +7,20 @@ import Base from './Base'
 const IFRAME_SRC = 'https://player.vimeo.com/video/'
 const MATCH_URL = /https?:\/\/(?:www\.|player\.)?vimeo.com\/(?:channels\/(?:\w+\/)?|groups\/([^\/]*)\/videos\/|album\/(\d+)\/video\/|video\/|)(\d+)(?:$|\/|\?)/
 const MATCH_MESSAGE_ORIGIN = /^https?:\/\/player.vimeo.com/
+const DEFAULT_IFRAMEPARAMS = {
```
CookPete: Can we call this `DEFAULT_IFRAME_PARAMS`?

src/players/YouTube.js
```diff
@@ -8,9 +8,17 @@ const SDK_URL = '//www.youtube.com/iframe_api'
 const SDK_GLOBAL = 'YT'
 const MATCH_URL = /^(?:https?:\/\/)?(?:www\.)?(?:youtu\.be\/|youtube\.com\/(?:embed\/|v\/|watch\?v=|watch\?.+&v=))((\w|-){11})(?:\S+)?$/
 const PLAYER_ID = 'youtube-player'
+const DEFAULT_PLAYERVARS = {
```
CookPete: `DEFAULT_PLAYER_VARS`

src/propTypes.js
```diff
@@ -6,6 +6,15 @@ export default {
   volume: PropTypes.number,
   width: PropTypes.oneOfType([ PropTypes.string, PropTypes.number ]),
   height: PropTypes.oneOfType([ PropTypes.string, PropTypes.number ]),
+  soundcloud: PropTypes.shape({
```
CookPete: I'm thinking we should perhaps use `soundcloudConfig`, `youtubeConfig` and `vimeoConfig` for clarity.

CookPete:
Typo: `iframParams` in commit message for 989d4f1076792fb747d86ba7a5cb912133da1638

CookPete:
Looking good Tim. I've made a few more notes, then we should be more or less good to merge :)

Fauntleroy:
Variable names changed, commit message amended.

CookPete:
Nice work. Could you squash the commits up into the different features you added? I think it could just be one for the config stuff, and one for the volume setting fixes. Leaving in fixup commits feels a bit untidy.

Fauntleroy:
Commits squashed.

Issue #7, opened 2015-11-01
YouTube player will not autoplay if window is not focused


Fauntleroy:
When you use a page with `react-player` in it, YouTube videos will not autoplay when the window does not have focus. While I'm certain YouTube has good intentions here, my application requires YouTube videos to play automatically regardless of whether the window is focused or not. Here are a few things I've tried but don't seem to work:

- Run `.playVideo()` after the `onReady` event of the YouTube player fires
- Constantly run `window.focus()` in the hope that they have a listener there (nope)

CookPete:
This seems to happen only when the SDK and iframe have not been loaded yet. That is, playing a youtube video straight after a youtube video works fine in the background because the player is already loaded and ready, but playing a youtube video after a Soundcloud track, or from a fresh load, will not work.

A possible (but annoying) solution would be to constantly have the youtube SDK and frame loaded as long as the `ReactPlayer` lives, but this wouldn't solve all the issues with this.

CookPete:
A quick [google search](https://www.google.co.uk/webhp?sourceid=chrome-instant&ion=1&espv=2&ie=UTF-8#q=chrome%20enable%20background%20autoplay) returns a few good results with more about this "feature". Notably a [post from Francois Beaufort](https://plus.google.com/+FrancoisBeaufort/posts/6PiJQqJzGqX) from Google, with various people complaining about this problem.

Fauntleroy:
According to google, if the tab has had focus at any point it should be able to autoplay media. In practice this is not the case. I need to build some test cases out and see if I can find out what the truth is.

Fauntleroy:
It looks like the issue is definitely related to players that use iframes.

My theory is that since the content served by the YouTube player's iframe is from a different domain, the tab focus flag is reset every time that player is destroyed and recreated. I'm going to figure out a way to keep the YouTube player's iframe around between player changes, which should eliminate this issue.

CookPete:
This would eliminate the issue, but we would still need an initial youtube play/click to "activate" the iframe, and I guess there would still be issues with playlists that don't begin with a youtube video. Either way, it sounds like the fix will not be very pleasant..

Fauntleroy:
Yup, you'll have to "prime" the Vimeo and YouTube players, but that shouldn't be *too* difficult. I'm gonna try my best not to make your codebase disgusting in the process :)

Fauntleroy:
[...]

PR #13, opened 2015-12-15
Prime players to enable autoplay when out of focus

CookPete:
My version of https://github.com/CookPete/react-player/pull/12 that attempts to fix [autoplay issues](https://github.com/CookPete/react-player/issues/7) when `ReactPlayer` is not in a focused tab. In short:

* Render all players at once, but only give `props` to the player that can play the current `url`
* Each player then sets `display: none` on itself if `url` is not present
* When `ReactPlayer` mounts, if `YouTube` or `Vimeo` players don't have a `url`, then `prime` the player by playing a short, silent video whilst hidden. This allows the players to autoplay any URLs given later on, even if not in a focused tab.

@Fauntleroy, any thoughts?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ All notable changes to this project will be documented in this file. This projec
 
 ### Unreleased
 
+* Fix [YouTube and Vimeo autoplay bug](https://github.com/CookPete/react-player/issues/7)
 * [Full commit list](https://github.com/CookPete/react-player/compare/v0.2.1...master)
 
 
--- a/README.md
+++ b/README.md
@@ -65,9 +65,13 @@ These props allow you to override the parameters for the various players
 
 Prop | Description
 ---- | -----------
-soundcloudConfig | An object containing configuration for the SoundCloud player. Includes `clientId`, which can be used to override the default `client_id`
-vimeoConfig | An object containing configuration for the Vimeo player. Includes `iframeParams`, which maps to the [parameters accepted by the Vimeo iframe player](https://developer.vimeo.com/player/embedding#universal-parameters)
-youtubeConfig | An object containing configuration for the YouTube player. Includes `playerVars`, which maps to the [parameters accepted by the YouTube iframe player](https://developers.google.com/youtube/player_parameters?playerVersion=HTML5)
+soundcloudConfig | Configuration object for the SoundCloud player. Set `clientId`, to your own SoundCloud app [client ID](https://soundcloud.com/you/apps)
+vimeoConfig | Configuration object for the Vimeo player. Set `iframeParams`, to override the [default params](https://developer.vimeo.com/player/embedding#universal-parameters). Set `preload` for [preloading](#preloading)
+youtubeConfig | Configuration object for the YouTube player. Set `playerVars`, to override the [default player vars](https://developers.google.com/youtube/player_parameters?playerVersion=HTML5). Set `preload` for [preloading](#preloading)
+
+##### Preloading
+
+Both `youtubeConfig` and `vimeoConfig` props can take a `preload` value. Setting this to `true` will play a short, silent video in the background when `ReactPlayer` first mounts. This fixes a [bug](https://github.com/CookPete/react-player/issues/7) where videos would not play when loaded in a background browser tab.
 
 ### Methods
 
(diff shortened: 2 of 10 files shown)


Fauntleroy:
I was thinking of making the priming functionality manual by using a configuration parameter:

```js
const reactPlayerProps = {
  youtubeConfig: {
    preload: true
  },
  vimeoConfig: {
    preload: true
  }
};
<ReactPlayer {...reactPlayerProps} />
```

This way we could avoid loading those APIs for those who don't intend on using them.

CookPete:
Updated. Moved priming logic from `Base` into `Youtube` and `Vimeo` players, so they can check their config and decide whether to prime or not. Should probably align the config name (currently `preload`) with the variable name (currently `priming`) and add some documentation. It also makes prop logic in `ReactPlayer` [rather awkward](https://github.com/CookPete/react-player/blob/fix-autoplay-prime/src/ReactPlayer.js#L52-L60) so there's potential for improvement there.

CookPete:
Updated. Now using slightly less ambiguous `preloading` var name

CookPete:
Updated (again) with a short explanation in `README.md`

CookPete:
Looking to merge this soon even if it's a bit weird.. @Fauntleroy are you happy enough?

Fauntleroy:
@CookPete I'll take a look at this soon to ensure that it works properly. My initial test earlier this week was having some problems. Can you check and see if `youtubeConfig` makes it all the way to the player?

CookPete:
> Can you check and see if youtubeConfig makes it all the way to the player?
 
Ah yeah that should have been sorted by https://github.com/CookPete/react-player/commit/52574f8aafffded6392016b8ede7b6c9d9d75fc1

Fauntleroy:
Something is definitely wrong when I try to use this in my application. I'm going to try a few things and see if I can resolve the problem.

Fauntleroy:
Autoplay seems to behave differently than it did before, but with `playing` and `youtubeConfig` i've managed to get that settled.

Volume, however, **is** broken. Looks like you just need to make sure volume is included in `shouldComponentUpdate`:

```js
shouldComponentUpdate (nextProps) {
    return (
      this.props.url !== nextProps.url ||
      this.props.playing !== nextProps.playing ||
      this.props.volume !== nextProps.volume
    )
  }
```

Fauntleroy:
I'm also running into `Uncaught TypeError: Cannot read property 'getFractionLoaded' of undefined` when using the SoundCloud player, though I'm trying to figure out how it's even happening.

Fauntleroy:
The issue with `getFractionLoaded` can be solved by checking to see if `this.refs.player` exists during the if statement wrapping the contents of `progress` in `ReactPlayer`:

```js
progress = () => {
    if (this.props.url && this.refs.player) {
      let progress = {}
      const loaded = this.refs.player.getFractionLoaded()
      const played = this.refs.player.getFractionPlayed()
```

Fauntleroy:
I'm also having issues rendering all of the players inside my application. For some reason only YouTube will bother to render. Here's what my React inspector is showing me:

screen shot 2015-12-18 at 11 55 52 pm

I imagine something horrible is happening to important parts of the code when minification occurs. Maybe all of the component names are being destroyed, and that's causing a few things to break here and there?

CookPete:
Wow, thanks very much @Fauntleroy for the testing! I've added [some more fixups](https://github.com/CookPete/react-player/compare/2dc18f7ef01341d84a7ff5951605106448643958...11b589d279b74776c0aa539cd832297e5d25636d) to hopefully solve most of the issues.

* Separate `load` and `play` methods are now used to clarify between actually *loading* media and *playing* it. Media will `play` after `load` if `props.playing` is true at the time. This should clear up the autoplay issues we've been seeing, and solidify the logic between loading and playing. For example, you can now keep `playing: false` but click through the demo and load media ready for playing whenever `playing` becomes `true.
  * This actually makes `autoplay` config overrides sort of obsolete/pointless. If you have `playing: true`, media should autoplay no matter what, and if `playing` is `false` then it shouldn't. Any other `autoplay` settings would just confuse the logic and the players' state.
* I've added the `volume` and `getFractionLoaded` fixes you suggested. Good spot!
* I need to look more at the minified problems. What exact compressor/options are you using?

Fauntleroy:
@CookPete it might not have to do with minification. I seem to remember being able to replicate the issue without minified code, but I can't remember exactly how that came about. I'll look into it more soon.

CookPete:
I'm gonna merge this in an attempt to avoid conflicts with other fixes. Hopefully all problems (other than the weird minified issue) have been sorted

Issue #20, opened 2016-01-02
Better handling of YouTube methods before player is ready


CookPete:
@Fauntleroy has found that `play()` can sometimes be called before `this.player.playVideo` exists. https://github.com/CookPete/react-player/commit/f18792a2da3167f775d48481aa8eae150b747d4d prevents errors for the time being, but doesn't really fix the problem.

From [this discussion](https://github.com/CookPete/react-player/commit/ef17dcc732d282f027aa3bddf8f91cbb05998cc8#commitcomment-15231248).

Fauntleroy:
@CookPete I think I found the culprit: `this.getSDK()` can get called more than once. It's initially called for the blank video load, but it will be called again if you load another video before the SDK loads.

PR #21, opened 2016-01-02
Fix YouTube onReady bugs

CookPete:
Attempts to fix https://github.com/CookPete/react-player/issues/20

@Fauntleroy I'm hoping this does the job? Could you confirm?

### DIFF
--- a/src/players/Base.js
+++ b/src/players/Base.js
@@ -38,7 +38,11 @@ export default class Base extends Component {
     if (this.props.playing || this.preloading) {
       this.preloading = false
       this.isReady = true
-      this.play()
+      if (this.loadOnReady) {
+        this.load(this.loadOnReady, this.props.playing)
+      } else {
+        this.play()
+      }
     }
   }
 }
--- a/src/players/YouTube.js
+++ b/src/players/YouTube.js
@@ -56,6 +56,11 @@ export default class YouTube extends Base {
       }
       return
     }
+    if (this.loadingSDK) {
+      this.loadOnReady = url
+      return
+    }
+    this.loadingSDK = true
     this.getSDK().then(YT => {
       this.player = new YT.Player(this.playerId, {
         width: '100%',
@@ -66,7 +71,10 @@ export default class YouTube extends Base {
           ...this.props.youtubeConfig.playerVars
         },
         events: {
-          onReady: this.onReady,
+          onReady: () => {
+            this.loadingSDK = false
+            this.onReady()
+          },
           onStateChange: this.onStateChange,
           onError: this.props.onError
         }


Fauntleroy:
Doesn't look like this fixes it... but I've got plenty of free time and a reproducible test case. I can go ahead and take care of this myself in a separate pull req if it's all the same to you

Fauntleroy:
I spent quite a while working on reproducing this bug, making a test for it, and resolving it—but I could only get one of those complete. Here are the steps I used to replicate my issue:

- The player must be initialized with `playing` set to `true` and `url` set to `null`
- The YouTube player must be set to `preload` in the `youtubeConfig`
- Immediately after the player initializes, the `url` must be set to another YouTube URL (I did this within the test app's `componentDidMount` call)

This will cause the player to attempt to prime the YouTube player with the empty video, thus calling `load`, which ultimately calls `new YT.player`. Immediately after that the `url` is updated, which causes `load` to be called again, calling `new YT.player` and overwriting its original value. Eventually the first player's `onReady` fires, but by that time the second player (which is still initializing) is the value of `this.player`. Thus we try to call `this.player.playVideo()` (and other methods) when one of these players has initialized, but the other hasn't.

My idea for fixing this would be to add a flag for keeping track of whether or not we've started initializing a YouTube player yet. Using this flag, you could check to see if a player is pending during a `load` call, then act accordingly. If no player is initializing, create a new YouTube player. If a player is initializing, add the video ID to a queue and fire `loadVideo` or `playVideo` when `onReady` fires.

CookPete:
>Eventually the first player's onReady fires

This PR should prevent that first `onReady` from firing.

CookPete:
Updated with a new approach. @Fauntleroy any better?

CookPete:
Argh, still not good enough. Pausing/playing is now screwed up.

CookPete:
Ah, never mind, I missed a `!` [here](https://github.com/CookPete/react-player/blob/master/src/players/YouTube.js#L84) which caused problems. Oops.. I've fixed up `master` and rebased this branch.

Fauntleroy:
This seems to be functioning as expected in both the react-player example and my application. I'm going to look over your code and see if there are any hidden gotchas before I sign off on this completely, though...

Fauntleroy:
Having trouble with this in production. Let me double check my work first

Fauntleroy:
Looks like the issue was probably caused by caching on my production server. Everything looks good to me here, though we should really start investing in tests as we move forward.

CookPete:
> we should really start investing in tests as we move forward.

Absolutely. Shallow rendering unit tests only get us so far. I'll open an issue.

I'll merge this and publish a patch release.

PR #30, opened 2016-01-19
Add elapsed and remaining time to demo

Conrad777:
Hey. First off, thank you for an awesome player component!!

I added elapses and remaining time displays to your demo. Thought it might help convince others searching for a player to use this one. 

**Note:** The countdown is a little jumpy every now and then. So it is entirely up to you if you want to add it or not, but I thought that it might be a good starting point for you to improve at a later stage.

### DIFF
--- a/src/demo/App.js
+++ b/src/demo/App.js
@@ -64,6 +64,19 @@ export default class App extends Component {
       </button>
     )
   };
+  addZero = number => {
+      return number > 9 ? "" + number: "0" + number;
+  };
+  convertToSeconds = (percentage) => {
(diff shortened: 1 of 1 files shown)


CookPete:
Hey @Conrad777, thanks for your PR. I actually have a small `Duration` component that I'm using in another app that uses `ReactPlayer`. I will probably make use of that to achieve the same effect here. It just means the code will be slightly less cluttered. Thanks for the effort though!

CookPete:
This is sorted in `v0.5.1`

PR #45, opened 2016-04-04
Add FilePlayer support for urls with query string

gridsane:
File urls could have a query string. For example: `vk.com` API returns the following urls:

```
https://psv4.vk.me/c611426/u238196536/audios/ccd84e1ca642.mp3?extra=HASH
```

### DIFF
--- a/src/players/FilePlayer.js
+++ b/src/players/FilePlayer.js
@@ -2,8 +2,8 @@ import React from 'react'
 
 import Base from './Base'
 
-const VIDEO_EXTENSIONS = /\.(mp4|og[gv]|webm)$/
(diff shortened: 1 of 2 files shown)


CookPete:
Nice job. The `dist` files don't need to be committed though. They are only for the sake of bower, and are built and added to the repo as part of the version bump script. Remove them from the commit and we are good to go!

gridsane:
Done.
Thanks for the awesome component!

CookPete:
Could you squash your new commit into the previous one?

gridsane:
Okay, done.

PS. Not sure if squash is visible in notifications.

CookPete:
Thanks! :+1:

PR #48, opened 2016-04-08
Pass on error event

### LINKED ISSUE #44: onError called without error message
I use the player similar to the demo App, including `onError={(e) => console.log('onError', e)}`.

onError is called each time the player is loaded, but the song plays just fine. The `e` is undefined.

Sometimes an mp3 URL can not be loaded from a remote source and the same error with an undefined error message is triggered.

Any idea how I can dig deeper into this to see what is going on?

pors:
The audio element onerror event was not passed on to the react-player onError method. This PR fixes that. This makes it possible to inspect the actual error: Event > srcElement:audio > error:MediaError > code [1,4]

Error codes can be found here: https://dev.w3.org/html5/spec-preview/media-elements.html#error-codes

This fixes #44

### DIFF
--- a/src/players/FilePlayer.js
+++ b/src/players/FilePlayer.js
@@ -16,7 +16,7 @@ export default class FilePlayer extends Base {
     this.player.onplay = this.onPlay
     this.player.onpause = () => this.props.onPause()
     this.player.onended = () => this.props.onEnded()
-    this.player.onerror = () => this.props.onError()
+    this.player.onerror = (e) => this.props.onError(e)
     super.componentDidMount()
   }
   load (url) {


CookPete:
I've been working on a more thorough fix for `onError` but if this helps you for now there's no reason not to merge it.

Could you remove the `dist` files from the commit? They are only there for `bower` and are automatically built and added to the repo when versioning.

CookPete:
Also @pors, could you also add a "Fixes #44" to the commit notes? So that `auto-changelog` will pick it up for the change log.

pors:
@CookPete all should be good now.

CookPete:
Merged this [manually](https://github.com/CookPete/react-player/commit/7995e869187c9c370bb4cb9a5dccb027ed7334fe) just to tidy up the commit message a bit. Nice work!

PR #65, opened 2016-05-20
SoundCloud to extend FilePlayer (and get rid of sc sdk)

denisflorkin:
Hello, from @CookPete suggestion I finally got somewhere with removing the need for sc sdk and removing most of the code from `SoundCloud`. The test passes (although only in chrome as I still haven't resolved my test env. issue...)

One weird thing is that I can get the sc stream into a `<audio>` but it does not work, and it works like a charm in a `<video>` I'm missing something somewhere...

Anyway not sure if My PR is really helping or if it's more a cry for a free lessons on react and classical inheritance....

Anyway here it is :)

### DIFF
--- a/src/players/FilePlayer.js
+++ b/src/players/FilePlayer.js
@@ -5,8 +5,6 @@ import Base from './Base'
 const VIDEO_EXTENSIONS = /\.(mp4|og[gv]|webm)($|\?)/
 const AUDIO_EXTENSIONS = /\.(mp3|wav)($|\?)/
 
-// const AUDIO_EXTENSIONS = /(\.mp3|\.wav|soundcloud.*)($|\?)/
-
 export default class FilePlayer extends Base {
   static displayName = 'FilePlayer'
   static canPlay (url) {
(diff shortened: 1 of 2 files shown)


CookPete:
I've been looking at this separately and have come up with a similar solution. The problem with your approach is that we lose the track poster image, which I would like to retain. It also makes me think that classic inheritance is not the clearest technique to use for the players, and maybe worth looking into composite components or something similar.

CookPete:
You can see a [branch of my attempt](https://github.com/CookPete/react-player/tree/minimal-soundcloud) with the main business happening in https://github.com/CookPete/react-player/commit/5393343a231ef4905271023c3e50a62fb8bec72e. However, I'm having trouble with getting the tests to run consistently, and I'm not sure if it's because of these changes or something else. Once I'm happy I'll merge it in to master.

CookPete:
https://github.com/CookPete/react-player/commit/5393343a231ef4905271023c3e50a62fb8bec72e is merged now so I'll close this off. Thanks for the PR!

PR #66, opened 2016-05-25
Update README.md links

raksonibs:


CookPete:
Weird. I don't remember leaving these empty.. Good spot :)

PR #69, opened 2016-06-01
Set playing to true by default

CookPete:
After some thought, I wonder why `playing` isn't `true` by default. Most users, especially those using `ReactPlayer` as a dumb player with no external controls, will want media to be playing if the component has been mounted. This saves a lot of unnecessary `playing={true}` props.

This may be a breaking change and perhaps should wait until `v1.0`, which I think should happen soon.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ import ReactPlayer from 'react-player'
 
 class App extends Component {
   render () {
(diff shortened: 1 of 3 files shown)


capaj:
@CookPete I am not so sure. Especially given that this breaks the whole component on mobile devices, I think it still makes more sense to have it explicit.

CookPete:
>Especially given that this breaks the whole component on mobile devices

This is a good point.

PR #70, opened 2016-06-01
Combine config into single config prop

CookPete:
I don't like how we have to carry around four separate `config` props in several places, especially for props that I think aren't used that often. This consolidates all the players' configuration data into one `config` prop, which falls back to a `defaultConfig` object for any property that isn't specified.

Still not sure if this is better than the current way of doing things. I'm happy to hear feedback if anyone has any.

This is definitely a breaking change and perhaps should wait until we're ready for a `v1.0` release.


CookPete:
Closing and opening a new PR because I messed up adding a milestone 😞

PR #87, opened 2016-07-26
Add support for non-es5 compatible browsers

dpiatek:
Hi @CookPete ! This adds two babel plugins that will make it possible to run non-dist code in non-es5 compatible environments. As far as I can tell this has no further effect on the codebase (tests passed fine locally).

http://babeljs.io/docs/plugins/transform-es3-member-expression-literals/
http://babeljs.io/docs/plugins/transform-es3-property-literals/

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -5,7 +5,11 @@
       "presets": ["react-hmre"]
     },
     "production": {
-      "plugins": ["add-module-exports"]
+      "plugins": [
+        "add-module-exports",
+        "transform-es3-member-expression-literals",
+        "transform-es3-property-literals"
+      ]
     }
   }
 }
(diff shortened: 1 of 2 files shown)


CookPete:
Hi @dpiatek, what is the reserved word that causes problems? I wonder if it will be cleaner to refactor the code to prevent the errors.

dpiatek:
The keyword in question is `default` and it's part of the module transpilation bit in `lib`.

CookPete:
Ah ok so it's a problem with babel's output itself, not the codebase. I've just read more about it [here](https://github.com/CookPete/react-player/pull/87). Thanks for the PR!

dpiatek:
thanks @CookPete - and for open sourcing this as well!

PR #95, opened 2016-08-10
Fix buildup of multiple message event listeners

jackyzhen:
When loading/reloading multiple vimeo players, there is a build up of message event listeners on the global window object, because there is no detach on unmount.


CookPete:
I've just realised this PR would have fixed https://github.com/CookPete/react-player/issues/101 which I fixed with a quick commit.. Apologies @jackyzhen! Nice work!

Issue #75, opened 2016-06-17
seekTo value and result mismatched


whitegrey:
If you use seekTo to set a value, the result may be wrong by 1s.

case in point:

seeking to 0:05 in a 148 second video means calling seekTo( 0.033783783783783786 ). onDuration will immediately return {played: 0.0337837775564004}. I'm assuming this is because React Player only sets to the 6th or 7th digit?

The problem is 0.0337837775564004 multiplied by 148s (video length) returns 4.999999078347259. Which youtube rounds down to 4s.

In other words, if you try to set time to 5s, you will instead get 4s.

On a side note, would you consider adding the options to set seek via seconds elapsed? It may just be me, but I feel it's a much more intuitive method of working with seek than percent played.

CookPete:
Your specific issue could be [YouTube adding one second to the reported duration](https://www.reddit.com/r/youtube/comments/4d4e4g/why_does_youtube_display_the_duration_of_a_video/) (possibly so videos between 0 and 1 second long appears as 1 second long). The reason fraction is used is because we only know the duration after `onDuration` is called, so the app would have to gather the duration into state (or somehow know it beforehand) before it is able to safely seek anywhere. With a fraction you can seek without needed to know the duration (because `0.5` will always be halfway through the media, etc).

I agree that is seems less intuitive, but unfortunately the different players make their durations available at different times during the loading sequence.

Perhaps `seekTo` could be clever and handle both. Seek using a fraction if the value given is less than 1, otherwise seek using seconds.

CookPete:
> So what happens when you do a seekTo(0.5) meaning you want to seek to 0.5 seconds ?

I did consider this when implementing but figured the chance of someone ever wanting to seek to `0.x` seconds was pretty much zero.

If you really wanted to (and the player SDK supported such specific seeking), you could use `getInternalPlayer()` and manually call `seekTo` (or whatever method is supported) on the internal player that ReactPlayer uses.

kalpesh4307:
@CookPete how can i use getInternalPlayer() and manually call seekTo in my video modal??

waynebloss:
Hello and thanks for your work on this player.

Everything is working great so far, but I really needed to be able to seek by `0.1` seconds for a video editor that I'm building. Overloading the `seekTo` method to do 2 separate functions by looking for fractions between '0' and '1' seems to go against best practices for making a clear and concise programming interface.

Would you consider a pull request that normalizes this behavior into separate functions?

[...]

PR #97, opened 2016-08-10
Add support for a `startOffset` (FilePlayer)

sveinburne:
The player will start playing at the given offset (in seconds).

**`startOffset` prop** : 
- default value is `0`
- is coerced (minimum `0`, maximum `player.duration`). Any forbidden value will be overwritten by default `0` (internal state), and a warning will be flushed to the console.
- is fully reactive: any update to a higher value then current playing cursor value will result in the update of the playing cursor value (`player.currentTime`).

**Following `FilePlayer` methods now use `startOffset` to compute their return values**
- `getDuration`
- `getFractionPlayed`
- `getFractionLoaded`

You can test this feature in the demo app !

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,6 +81,7 @@ Prop | Description | Default
 `className` | Pass in a `className` to set on the root element
 `style` | Add [inline styles](https://facebook.github.io/react/tips/inline-styles.html) to the root element
 `progressFrequency` | The time between `onProgress` callbacks, in milliseconds | `1000`
+`startOffset` | The offset at which the player will start, in seconds. Works with FilePlayer only. | `0`
 
 #### Callback props
 
--- a/src/ReactPlayer.js
+++ b/src/ReactPlayer.js
@@ -31,7 +31,8 @@ export default class ReactPlayer extends Component {
       this.props.playing !== nextProps.playing ||
       this.props.volume !== nextProps.volume ||
       this.props.height !== nextProps.height ||
-      this.props.width !== nextProps.width
+      this.props.width !== nextProps.width ||
+      this.props.startOffset !== nextProps.startOffset
     )
   }
   seekTo = fraction => {
--- a/src/demo/App.js
+++ b/src/demo/App.js
@@ -16,7 +16,8 @@ export default class App extends Component {
     volume: 0.8,
     played: 0,
     loaded: 0,
-    duration: 0
+    duration: 0,
+    startOffset: 0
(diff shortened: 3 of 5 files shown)


CookPete:
This is a lot of complex code (and wierd use of state) for such a specific use case (and something that you can achieve with logic outside of `ReactPlayer`) so I am going to close this off. If you feel like `startOffset` is an important possible feature then feel free to open an issue and we can discuss it over there. I would rather enable the prop for all players and try to avoid using state inside the player classes.

All you would need to do to achieve this outside of `ReactPlayer` is (assuming your player has a ref of `player`):

* Listen to the `onPlay` prop and then immediately run `this.refs.player.seekTo(fraction)`
* Amend the amounts returned from `onProgress` and `onDuration` before using them to update your UI, according to the fraction of the media you want to play

sveinburne:
The reason why internal state shall be used is that you know the duration after meta data has been loaded. This is mandatory if you want to have a bounded, consistent start offset. Plus, `startOffset` is a prop, which makes it inherently dynamic. You should handle props value updates to expose a consistent behaviour. 

I don't find anything complex, except using `startOffset` when computing duration, fraction played and fraction loaded. The `componentWillReceiveProps` code handles `startOffset` prop updates safely (bounded within [`0`, `player.duration`]) and drags the cursor (`currentTime`) to the beginning of the offset to avoid a virtually negative cursor (`player.currentTime < startOffset`).

`startOffset` state value is the safe counterpart of the `startOffset` prop. I could refactor to make all this logic clearer.

However I do understand that you would like all players to implement this feature if it shall be published.

What is the complexity you are describing and do you find it still accurate after this reading?

CookPete:
I think the same thing can be achieved for all players with less intrusive code. The logic relies on the `metadataloaded` event which we would have to ensure works in all browsers. If we need to wait until we know the duration, we could maybe latch on to the existing `onDuration` logic. Also, the component's methods currently consistently use a `fraction` decimal instead of `seconds` for stuff like `seekTo`, and the `onProgress` callback. Agreed that this [should probably be seconds](https://github.com/CookPete/react-player/issues/75) but I would rather change the paradigm across the whole component than add a new feature that uses seconds and doesn't follow the pattern of other props/methods.

I'm happy to open an issue requesting this feature and to work out a clean implementation if it gains support from other devs, but I don't want to add lots of code for a single feature that solves a rather edgy use case.

PR #102, opened 2016-09-01
Add `hidden` prop to ReactPlayer

kevinzwh:
**Description:**
`hidden` is a standard HTML prop of react. I was building a toggle feature to hide ReactPlayer -- as opposed to re-rendering the player each toggle, which would lag due to having to retrieve and re-embed the media each time -- and found that ReactPlayer does not accept a hidden prop to hide its top level div. In order to hide it using `hidden`, I had to write another parent div above ReactPlayer just to hide the component.

```es6
// Not the best way to hide a component
<div hidden={ myBoolean }>
  <ReactPlayer url={ this.props.randomURL }>
</div>
```

**Changes Proposed:** 
I'd like to propose that `hidden` be added to one of the standard props ReactPlayer accepts (alongside the standard `style` & `className` props). Like so:

```es6
// Feels better this way!
<ReactPlayer
  url={ this.props.randomURL }
  hidden={ myBoolean }
/>
```

I've included working code that allows this very use case in this PR. I've also included a line to the `Props` documentation in README.md about `hidden` as a prop that can hide ReactPlayer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,6 +78,7 @@ Prop | Description | Default
 `volume` | Sets the volume of the appropriate player | `0.8`
 `width` | Sets the width of the player | `640`
 `height` | Sets the height of the player | `360`
+`hidden` | Set to `true` to hide the player | `false`
 `className` | Pass in a `className` to set on the root element
 `style` | Add [inline styles](https://facebook.github.io/react/tips/inline-styles.html) to the root element
 `progressFrequency` | The time between `onProgress` callbacks, in milliseconds | `1000`
--- a/src/ReactPlayer.js
+++ b/src/ReactPlayer.js
@@ -94,7 +94,7 @@ export default class ReactPlayer extends Component {
       height: this.props.height
     }
     return (
-      <div style={style} className={this.props.className}>
+      <div style={style} className={this.props.className} hidden={this.props.hidden}>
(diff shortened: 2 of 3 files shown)


kevinzwh:
To verify full functionality that `hidden` works, you can try modifying the demo as such:

```es6
//src/demo/App.js
//...
//...
export default class App extends Component {
  componentDidMount() {
    var interval = setInterval(this.toggleHide, 1000); // Hide the player every 1000ms
  }
  toggleHide = () => this.setState({ hidden: !this.state.hidden }) // Toggle hidden state
//...
//...
  render () {
//...
//...
    return (
      <div className='app'>
        <section className='section'>
          <h1>ReactPlayer Demo</h1>
          <ReactPlayer
            ref='player'
            className='react-player'
            width={480}
            height={270}
            hidden={this.state.hidden} // Provide hidden state as a prop to ReactPlayer
//...
//...
```

kevinzwh:
Thoughts on this, @CookPete?

CookPete:
Yeah this looks good. I would normally have just used `className` or `style` to hide the player but I wasn't aware of the `hidden` attribute. Nice work!

As this is adding new functionality, I'll publish as a new minor version. Should probably look at moving to `1.0.0` soon.

PR #111, opened 2016-10-05
Pass DOM node to YouTube Player instead of ID

jsibelman:
Our use case involves loading the player as part of a component in an iframe, so using getElementById fails. Changing it to use the node has the side benefit of making the component a little more flexible, since now the id isn't required, and subsequently there's no need for tracking unique ids, and it avoids id collisions on the outside chance someone might not be cautious about their naming.


CookPete:
I didn't know you could do this. Much better! Thanks for the PR.

PR #118, opened 2016-10-27
add iframe style for vimeo player

mzbac:
this PR aim to accept style setting from IframeParams, apply iframe style. it will allow user to specify unfixed width and height of player container element (this will remove the black side bar in player container). use case as below:

``` js
 <div style={{ width: '100%', height: '0', position: 'relative', paddingBottom: '56.25%' }} >
        <iframe
          style={{ width: '100%', height: '100%', position: 'absolute', top: '0', left: '0' }}
```

### DIFF
--- a/src/players/Vimeo.js
+++ b/src/players/Vimeo.js
@@ -107,12 +107,12 @@ export default class Vimeo extends Base {
     this.iframe = iframe
   }
   render () {
-    const { fullscreen } = this.getIframeParams()
+    const { fullscreen, iframeStyle = {} } = this.getIframeParams()
     const style = {
       display: this.props.url ? 'block' : 'none',
       width: '100%',
-      height: '100%'
-    }
+      height: '100%',
+      ...iframeStyle}
     return (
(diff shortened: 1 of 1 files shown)


CookPete:
You can do this with a wrapper around ReactPlayer, like in [rplayr](http://rplayr.com/r/chillmusic):

```js
<div style={{ position: 'relative', paddingBottom: '56.25%' }}>
  <ReactPlayer
    url={url}
    width='100%'
    height='100%'
    style={{ position: 'absolute', top: 0, left: 0 }}
  />
</div>
```

Also see the [relevant code from rplayr source](https://github.com/CookPete/rplayr/blob/master/src/components/Player.scss#L4-L21), although this uses classnames instead of inline styles.

CookPete:
Here's a [working jsFiddle](https://jsfiddle.net/bce4a2pv/) too.

mzbac:
@CookPete  thanks for your pointer. I will close this PR since the issue has been solved

Issue #126, opened 2016-11-20
Possible bug onProgress


tarhan:
Currently **onProgress** reporting only if ```playing``` prop set to ```true```.  

In result if I allow Youtube's own control by setting ```controls``` and do not set ```playing``` to ```true```. Even if I do not set ```controls``` I still can resume Youtube video and I do not receive progress updated.  
I'm not sure if you decide to use such behavior intentionally or not. If not I supplied simple fix.

In my opinion it is desirable to received updates about playing progress despite cause it changed: due React actions or due by embed player itself.

tarhan:
I've just found about #51 and #94. So I understand that you disabled progress reporting intentionally.  
But in some cases like mine I want embedded player to be able to fully control playing and I only received its events. So I want to receive progress monitoring to correctly show external subtitles below video.  
Maybe it is worth to add additional bool prop like ***reportProgress***. Then user decide by himself to bind ```playing``` and ```reportProgress``` together or to use them separately.

CookPete:
>Maybe it is worth to add additional bool prop like reportProgress. Then user decide by himself to bind playing and reportProgress together or to use them separately.

I think it will be easier to simply always report `played` in the `onProgress` callback, and (like you said) users must decide whether to update their UI based on their own `playing` state. 

We could even force an `onProgess` call before any `pause` happens, to avoid the problem outlined in https://github.com/CookPete/react-player/issues/51.

The real solution to your problem would be to store your own `playing` state that gets updated using the `onPlay` and `onPause` callbacks. If you are just setting `playing={false}` to a player that actually could be playing, you are exposing yourself to more issues like this in future.

tarhan:
> I think it will be easier to simply always report played in the onProgress callback, and (like you said) users must decide whether to update their UI based on their own playing state.

So it is like in my PR. I'm always reporting ```played``` position.

> The real solution to your problem would be to store your own playing state that gets updated using the onPlay and onPause callbacks. If you are just setting playing={false} to a player that actually could be playing, you are exposing yourself to more issues like this in future.

I'm not setting ```playing={false}```. I'm just leaving it out and of course it have same result. And most players Youtube, DailyMotion, Tudou and many others even if their support disabling player controls still able to pause/resume video by clicking on video.  
So either to force user to always track onPlay/onPause/playing and receive progress updates or to always receive progress updates.

CookPete:
https://github.com/CookPete/react-player/pull/127 is now merged and published in `0.13.0` so this should be sorted now.

PR #127, opened 2016-11-20
Fix onProgress non-reporting if playing prop is not set.

tarhan:
Simple fix for onProgress bug mentioned here: #126


CookPete:
Why change the `onProgress` default to `null` and add a guard against it? Seems like an attempt to solve a problem that doesn't exist, and isn't related to the commit.

CookPete:
Fixed up and merged manually. Thanks!

PR #136, opened 2016-12-13
Update dependencie node sass

felquis:
I used npm-check-updater and node-sass was outdated.

I also added a yarn.lock since its a good tools for those who need caching and lock files.


felquis:
I'd suggest to not use ^ in dependencies, since sometimes modules can break from patches, what you think about it? Should I send another PR removing it?

felquis:
Most of the tests failed in my machine too, not from section YouTube Firefox 49.0.0 (Mac OS X 10.10.0), 

```log
CSS2Properties doesn't have an indexed property setter for '0'
      CSSPropertyOperations.setValueForStyles@webpack:///~/react-dom/lib/CSSPropertyOperations.js:190:0 <- karma.webpack.js:12557:10
      ReactDOMComponent.Mixin._updateDOMProperties@webpack:///~/react-dom/lib/ReactDOMComponent.js:874:0 <- karma.webpack.js:12182:8
      ReactDOMComponent.Mixin.mountComponent@webpack:///~/react-dom/lib/ReactDOMComponent.js:514:0 <- karma.webpack.js:11822:8

```

CookPete:
>I used npm-check-updater and node-sass was outdated.

You have bumped the version but added it to `dependencies`, even though it belongs in `devDependencies` as it is only used for building the demo page.

>I'd suggest to not use ^ in dependencies, since sometimes modules can break from patches, what you think about it? Should I send another PR removing it?

This seems redundant if we add a `yarn.lock` file. People using `yarn` will get the locked in dependency versions and those using `npm` will get the latest packages and bear the risk of something breaking, although I think generally we should trust package owners to use semver correctly and not release breaking changes in `patch` or `minor` version bumps.

>Most of the tests failed in my machine too, not from section YouTube Firefox 49.0.0 (Mac OS X 10.10.0),

The tests aren't 100% reliable on all machines, hence why I [removed them from `preversion`](https://github.com/CookPete/react-player/commit/ed6cab0cc983efaf0ef8813b77a0a3a407091726#diff-b9cfc7f2cdf78a7f4b91a753d10865a2). I just ran them with the latest browser versions and they all passed.

Issue #140, opened 2017-01-09
Callback conflict with multiple Vimeo ReactPlayers on the same page


nikolas:
### Expected Behavior
The onProgress and onDuration callbacks for different ReactPlayer instances don't interact with eachother.

### Actual Behavior
Any onProgress or onDuration callback gets called when multiple Vimeo ReactPlayers are on the page.

### Steps to Reproduce
Render these two components to a page:
```js
export default class PlayerA extends React.Component {
  render() {
    const url = 'https://vimeo.com/189968214';
    return <ReactPlayer url={url} onDuration={this.onDuration} />;
  }
  onDuration(duration) {
    console.log('onDuration A', duration);
  }
}

export default class PlayerB extends React.Component {
  render() {
    const url = 'https://vimeo.com/196277011';
    return <ReactPlayer url={url} onDuration={this.onDuration} />;
  }
  onDuration(duration) {
    console.log('onDuration B', duration);
  }
}
```
"onDuration A" gets called twice, once correctly, and once with the incorrect duration, coming from PlayerB.

pwbrown:
On a similar note I am also experiencing the "onPlay" event being called on one video when another video has started playing.  This does not happen with Youtube videos, but as soon as I switched to Vimeo the interaction between multiple videos on a page starts colliding.

CookPete:
Hopefully this is fixed by [migrating to the new Vimeo API](https://github.com/CookPete/react-player/commit/f262c1e87be887d06c01217bc86a6f55dc90085e). If not, feel free to reopen.

Issue #90, opened 2016-08-02
Migrate to new Vimeo API


CookPete:
There is a new [Vimeo Player API](https://github.com/vimeo/player.js) which looks much better than the current API, so `ReactPlayer` should probably start using that.

manpenaloza:
I do some client stuff with that new Vimeo API at the moment. Is this something still pending here? If yes, I can try making some progress here.

CookPete:
I already have a working branch so don't worry too much. I just haven't got around to tidying it up and merging yet. The only remaining thing is an [issue](https://github.com/vimeo/player.js/issues/38) with player dimensions, but I think I have a workaround.

manpenaloza:
ok, in case you need support on this, just ping me.

nikolas:
@CookPete can you make your new Vimeo API branch public? I can't see it listed under branches. I'd like to start using with the hope that it will address this Vimeo issue I ran into today: https://github.com/CookPete/react-player/issues/140

pwbrown:
Thanks @CookPete for the public vimeo branch

CookPete:
Fixed by https://github.com/CookPete/react-player/commit/f262c1e87be887d06c01217bc86a6f55dc90085e

PR #142, opened 2017-01-17
Use new Vimeo player.js API

nikolas:
These changes turn out to solve my big problem I'm having with issue #140.

Also addresses issue #90.

I'm sure there are things that are missing from here. I realize you have your own vimeo branch going - I'm putting this up here for reference, to share thoughts, etc., and because it solves my problem.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,4 +1,2 @@
-src
 test
 server.js
-webpack.config.*.js
--- a/package.json
+++ b/package.json
@@ -92,6 +92,7 @@
     "whatwg-fetch": "^2.0.1"
   },
   "dependencies": {
+    "@vimeo/player": "^2.0.1",
     "fetch-jsonp": "^1.0.2",
     "load-script": "^1.0.0",
     "query-string": "^4.2.3"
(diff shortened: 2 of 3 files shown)


nikolas:
Hmmm.. tests are passing for me locally (they're run in Firefox and Chromium). Looking into why travis is failing.

CookPete:
Nice work @nikolas. My only issue is that I don't like the idea of directly `import`ing the vimeo package into the `ReactPlayer` library, which needlessly increases the package size even if the vimeo player is not used. This is why I load the vimeo SDK in using `load-script` in my branch (same with the youtube SDK). 

Is there anything different you have done compared to my attempt that I should keep in mind? What I'll probably end up doing is combining the wisdom of both before merging it in.

nikolas:
Ah okay - good point with importing the vimeo package. I was wondering why you didn't do that with your branch and have the loadSDK stuff, but now I get it.

CookPete:
Yeah I wanted to avoid a massive `ReactPlayer` bundle with every single SDK embedded into it, considering the majority of users will only use it for one or two types of URL (possibly just file paths, which uses no SDK at all).

PR #151, opened 2017-02-12
fixed width & height of video when it is in fullscreen mode

nndung179:
When video is at a fullscreen mode at demo page. The video is not actually in fullscreen mode. Because the width the height passed in from the props. So that I changed width from 480 x 270 => 100% x 100% and wrapped outside a div (480x270). Now it is worked fine.


iam-peekay:
@nndung179 excellent. yeah was going to suggest this too

PR #176, opened 2017-03-29
The *playsinline* attribute doesn't need the *webkit* prefix on iOS 10.2

laustdeleuran:
Inline videos no longer work on iOS 10.2, because Mobile Safari no longer seems to support `webkit-playsinline` but `playsinline` instead. Added the latter and kept the former in the code for maximum support. 

Thanks for a great lib!


CookPete:
Hopefully https://github.com/CookPete/react-player/commit/39a338b690aaacb3de3b38afef9016f240b4abf9 solves this. I added a top level `playsinline` prop a while ago but never pushed it. Apologies! Thanks for the PR.

Issue #167, opened 2017-03-09
Callbacks for click, hover, keydown


oyeanuj:
Hey @CookPete! So far, things have been going great while integrating with `react-player` but ran into a small roadblock. I was trying to implement - 

1. Play/Pause `onClick`
2. Show/hide controls `onMouseOver`/`onMouseOut`
3. Keyboard manipulations `onKeyDown`

But it seems none of these events are exposed at the moment. My current workaround is to have a `div` around `ReactPlayer` and tying all the callbacks to that `div` but it isn't the nicest solution.

Thoughts? Do you think these can be exposed?

Thank you!

CookPete:
Yeah an object of props to attach to the wrapper div makes sense. What name would make sense though? `events`?

```js
<ReactPlayer
  url='file.mp3'
  events={{
    onClick: () => {},
    onMouseOver: () => {}
  }}
/>
```

Maybe `events` doesn't make sense as it could hold _any_ attribute to attach to the wrapper. Maybe `wrapperProps` or `wrapperAttributes`? Is there an established term for "other props to blindly attach to a component"?

oyeanuj:
In my experience, I think most components just pass through any other props with just the spread syntax after slicing off the props they needed.

So, I feel like we might not even need a wrapper unless you feel it is cleaner.

CookPete:
There is already a wrapper div [here](https://github.com/CookPete/react-player/blob/master/src/ReactPlayer.js#L108) that we can attach stuff to. The problem is that using the spread syntax would require something like
```js
const { url, playing, volume, blah, blah, blah, ...extra } = this.props
```
which would be huge and impractical considering the amount of [props](https://github.com/CookPete/react-player/blob/master/src/props.js) we have. That's why I'm thinking something like `wrapperProps` or maybe just `attributes` so that we can simply do
```jsx
<div style={...} className={...} {...attributes} />
```
for the wrapper.

PR #183, opened 2017-04-14
Apply all non-valid props to the wrapper element

CookPete:
Possible fix for https://github.com/CookPete/react-player/issues/167

@oyeanuj any thoughts on this? The only possible issue is that we now call lodash's `omit` on every render, but I'm not sure it will affect performance that badly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,8 +83,6 @@ Prop | Description | Default
 `playbackRate` | Sets the playback rate of the appropriate player | `1`
 `width` | Sets the width of the player | `640`
 `height` | Sets the height of the player | `360`
-`hidden` | Set to `true` to hide the player | `false`
-`className` | Pass in a `className` to set on the root element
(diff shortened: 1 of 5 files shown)


oyeanuj:
Hey @CookPete! The lodash omit part looks good, I'd be surprised if it affects performance. I did have one question though - we are attaching the events/props to the `div` wrapper which can enclose multiple players (if I am correct in the understanding of how `renderPlayers` function is working. 

I feel like for the original case I mentioned in #167, one would need these controls on individual players - like hovering on an individual player should only show that player's controls. 

Thoughts?

CookPete:
>we are attaching the events/props to the div wrapper which can enclose multiple players (if I am correct in the understanding of how renderPlayers function is working.

Multiple players are only ever rendered when `preload` is set and youtube or vimeo players need to be "primed" by starting in the background. These players are never displayed so it doesn't make a difference.

oyeanuj:
Oh, I see. Well, since they don't show up visually, then all of this looks good! Thanks so much!

PR #185, opened 2017-04-27
Adding more changes made to the log

oyeanuj:
Hi @CookPete, I noticed that there were a few more changes made that weren't listed on the Changelog. So, just a little PR to add that :)

Thanks for all the updates!


CookPete:
Hey @oyeanuj, thanks for the PR.

Unfortunately I am using a beast of my own creation, [`auto-changelog`](https://github.com/CookPete/auto-changelog), to automatically generate the changelog when new versions are released. Unfortunately it has failed pretty badly for this latest version, and it isn't clever enough to keep any manual edits that are added after generation. The proper thing to do is to fix `auto-changelog` to include more notes from the commits and merges, and then regenerate the changelog.

Thanks for pointing it out!

oyeanuj:
Ah, got it! That seems like a pretty fancy tool.. good to know about it!

PR #187, opened 2017-05-05
Fix initial loading bug for Private Vimeo videos

quicksnap:
Private Vimeo videos have the structure `https://vimeo.com/<number>/<hex>`. For some reason, when passing just the ID to the initial player, it chokes with errors. When the player is already loaded, using the ID seems to work fine.


CookPete:
Thanks!

CookPete:
Published in `0.17.1`

quicksnap:
Wow! Thanks for publishing so quickly!

PR #199, opened 2017-05-29
Soundcloud: use Widget API instead of HTTP API

### LINKED ISSUE #181: Soundcloud client ID request rate limit reached
Following on from https://github.com/CookPete/react-player/issues/170, this library has become popular enough to be consistently hitting the request rate limit using the built-in Soundcloud client ID, meaning Soundcloud tracks will not work on any players that have not been configured with their own ID. 

Perhaps in `v1.0.0` the library should force users to pass in their own client IDs.

xouabita:
Using the Soundcloud Widget API doesn't require a client id, so this will fix #181.

Info about Widget API: https://developers.soundcloud.com/docs/api/html5-widget

### DIFF
--- a/src/players/SoundCloud.js
+++ b/src/players/SoundCloud.js
@@ -63,6 +63,13 @@ export default class SoundCloud extends Base {
       player.play()
     })
   }
+  getSongData () {
+    return new Promise(resolve => {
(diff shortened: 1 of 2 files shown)


CookPete:
Manually merged. Thanks for the PR!

PR #201, opened 2017-05-30
More accurate played fraction for Wistia

rijk:
I noticed that the Wistia `percentWatched()` is actually pretty rough; it's only updated in chunks, which sucks if you need frame precision. To fix this, I’ve reimplemented the `getFractionPlayed()` calculation to use the Wistia `time()` and `duration()` methods, which are always 100% accurate.


rijk:
Any chance this one can be merged?

CookPete:
I intended to merge this as part of `0.19.0` but must have missed it. Just published in `0.19.1`. Thanks! 🎉

PR #216, opened 2017-07-26
allow preload in default player

adalbertoteixeira:
The default player had `auto` hardcoded. This PR allows the `preload` property to be passed into the player, defaulting to `auto`.


CookPete:
You can already override this using `fileConfig.attributes`:

```jsx
<ReactPlayer
  playing
  url='file.mp4'
  fileConfig={{ attributes: { preload: 'whatever' }}}
/>
```

Adding a top level `preload` prop for the sake of one player seems like overkill.

PR #230, opened 2017-09-04
Fixed broken TypeScript type definitions

### LINKED ISSUE #229: TypeScript types are broken after last updates
Compiling with TypeScript is actually not possible
```
Using tsc v2.4.2
node_modules/react-player/index.d.ts(32,4): error TS1005: ';' expected.
node_modules/react-player/index.d.ts(32,5): error TS1131: Property or signature expected.
node_modules/react-player/index.d.ts(39,4): error TS1005: ';' expected.
node_modules/react-player/index.d.ts(39,5): error TS1131: Property or signature expected.
node_modules/react-player/index.d.ts(46,4): error TS1005: ';' expected.
node_modules/react-player/index.d.ts(46,5): error TS1131: Property or signature expected.
node_modules/react-player/index.d.ts(49,4): error TS1005: ';' expected.
node_modules/react-player/index.d.ts(49,5): error TS1131: Property or signature expected.
```
PR follows

k7sleeper:
Fix #229

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,11 +1,11 @@
 import * as React from 'react';
 
-export interface SourceSpec {
+export interface SourceProps {
   src: string;
   type: string;
 }
 
-export interface TrackSpec {
+export interface TrackProps {
   kind: string;
   src: string;
   srcLang: string;
@@ -13,7 +13,7 @@ export interface TrackSpec {
 }
 
 export interface ReactPlayerProps {
-  url?: string|string[]|SourceSpec[];
+  url?: string|string[]|SourceProps[];
   playing?: boolean;
   loop?: boolean;
   controls?: boolean;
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -1,7 +1,19 @@
 import * as React from 'react';
 
+export interface SourceSpec {
```
CookPete: Is there any point `export` ing `SourceSpec` and `TrackSpec`? They are only used in this file.

index.d.ts
```diff
@@ -1,7 +1,19 @@
...
+  src: string;
+  type: string;
+}
+
+export interface TrackSpec {
```
CookPete: I think the names `SourceProps` and `TrackProps` make more sense?

index.d.ts
k7sleeper: > They are only used in this file.
Sure?
If you want to prepare an object of type `SourceSpec` somewhere in your source you'll be happy to be able to import that type.

index.d.ts
k7sleeper: +1

PR #234, opened 2017-09-12
update of DOM listeners in FilePlayer when audio/video tags was switched

Phationmationion:
Hey!
Big thanks for your work! I use your player in my project and it works good and simply. But when I need to switch videofile to audiofile, all used DOM-listeners gets old and it crashes player normal work.
So... Hope my commit can make your player a little bit better :)

### DIFF
--- a/src/players/FilePlayer.js
+++ b/src/players/FilePlayer.js
@@ -34,31 +34,20 @@ export default class FilePlayer extends Base {
     )
   }
   componentDidMount () {
-    const { playsinline, onPause, onEnded, onError } = this.props
-    this.player.addEventListener('canplay', this.onReady)
-    this.player.addEventListener('play', this.onPlay)
-    this.player.addEventListener('pause', () => {
-      if (this.mounted) {
-        onPause()
-      }
-    })
-    this.player.addEventListener('seeked', this.onSeek)
-    this.player.addEventListener('ended', onEnded)
-    this.player.addEventListener('error', onError)
-    if (playsinline) {
-      this.player.setAttribute('playsinline', '')
-      this.player.setAttribute('webkit-playsinline', '')
-    }
+    this.addDOMListeners()
     super.componentDidMount()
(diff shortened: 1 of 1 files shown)


src/players/FilePlayer.js
```diff
@@ -34,31 +34,20 @@ export default class FilePlayer extends Base {
...
-    }
+    this.addDOMListeners()
     super.componentDidMount()
   }
+  componentDidUpdate (prevProps) {
+    const { url, config } = this.props
+    const wasAudio = AUDIO_EXTENSIONS.test(prevProps.url) || prevProps.config.file.forceAudio
+    const isAudio = AUDIO_EXTENSIONS.test(url) || config.file.forceAudio
+    if (wasAudio !== isAudio) {
+      this.removeDOMListeners()
```
CookPete: I'm not sure this is necessary. At this point, `this.player` is already the new `<audio>` or `<video>` element and the old one has been unmounted and removed from the DOM. We might have to do the same check in `componentWillReceiveProps` and run `removeDOMListeners` whilst the old element is still mounted? Or we might not have to run it at all, as the element is being killed anyway.

Phationmationion: Yes, I was trying to set this check in cwrp, but by unknown (for me) reasons it don't works: player crashes. And by some unknown reasons, when we skip removing listeners and replace one tag by another, method getEventListeners() in console of my Chrome shows that listeners exists on fresh tag too. Possibly react make this magic, possibly it such specifical Chrome (devTools?) bug. To be honest, I did not undestand the problem deep, but in PR I provided the only working way I found.

CookPete: I'm happy to merge the PR and will have a look at fixing it up a bit. Thanks!

Phationmationion: super.componentWillReceiveProps(nextProps)! Exactly!

PR #235, opened 2017-09-14
Updated TypeScript typings

Tenga:
This PR will update the TS typings, as they seem to have gone out of date with the 2.4 config changes.

I've made each property in each config optional, as from what I can tell, most of them have defaults and are not mandatory. It also allows the authors to pass in partial objects (file config with just `attributes` instead of forcing you to define `tracks`, `forceAudio`, etc.)

Each individual config for each video provider are also now their own exported interfaces so you can reuse the types without having to recreate them yourself in cases like this:

```TS
import ReactPlayer, { FileConfig, TrackProps } from "react-player";

// ...
const getTracks = (): TrackProps[] => {
  // Whatever get tracks does...
}

// ...
const fileConfig: FileConfig = {
  tracks: getTracks(),
}

// ...
<ReactPlayer config={{ file: fileConfig }} />
```

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -12,52 +12,80 @@ export interface TrackProps {
   default?: boolean;
 }
 
+export interface SoundCloudConfig {
+  options?: Object;
+}
+
+export interface YouTubeConfig {
+  playerVars?: Object;
+  preload?: boolean;
+}
+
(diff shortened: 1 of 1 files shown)


CookPete:
Nice work 👍

PR #240, opened 2017-10-02
Fix Vimeo instantiation error catching

rijk:
Errors happening during instantiation of the Vimeo player (for example: a private video or 404) were not caught and passed to the error handler. This resulted in a bunch of console errors (triggered by the `this.player.on(...)` calls after instantiation), a broken player, and no way to do anything about it or even tell the user what is going on.

This change calls `player.ready()` right after instantiation (which returns a [promise](https://github.com/vimeo/player.js/blob/master/src/lib/embed.js#L52)), allowing us to check if everything is OK (in which case we attach all the listeners), or not (in which case we pass the error on to our `onError` method).

More info: https://github.com/vimeo/player.js/issues/165

### DIFF
--- a/src/players/Vimeo.js
+++ b/src/players/Vimeo.js
@@ -39,27 +39,29 @@ export default class Vimeo extends Base {
         url,
         loop: this.props.loop
       })
-      this.player.on('loaded', () => {
-        this.onReady()
-        const iframe = this.container.querySelector('iframe')
-        iframe.style.width = '100%'
-        iframe.style.height = '100%'
-      })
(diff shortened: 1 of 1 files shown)


CookPete:
https://github.com/CookPete/react-player/commit/d34d2286239bbf5caf1236fa4f6fa58bcb8a801e should fix this, but there is still a bug with the Vimeo API that renders the player useless after the initial error occurs: https://github.com/vimeo/player.js/issues/166.

Thanks for the PR!

PR #248, opened 2017-10-12
Add crossorigin to propTypes

btnwtn:
Support `crossOrigin` prop on the video element.


CookPete:
The top level props are used for ReactPlayer functionality, not specifically for the `<video>` element.

To apply attributes to the `<video>` element use `config.file.attributes`:

```jsx
<ReactPlayer
  url='file.mp4'
  config={{ file: { attributes: { crossOrigin: true } } }}
/>

Issue #250, opened 2017-10-16
Vimeo duration not available until starting playback


rijk:
To reproduce:

1. Go to http://cookpete.com/react-player/
2. Click Stop button to prevent auto playing video
3. Click Test A button for Vimeo

Expected behaviour:

`duration | 1:00`

Actual behaviour:

`duration | 0:00`

This is a problem especially when trying to implement https://alexanderwallin.github.io/react-player-controls/#ProgressBar. Because there is no known `totalTime`, the bar produces an error upon seeking (`Seconds must be a positive number less than the duration of the video`).

rijk:
This is my workaround for the moment:

```javascript
if ( isVimeoUrl( videoUrl ) ) {
	this.player.getInternalPlayer().getDuration().then(
		duration => this.setState({ duration })
	)
}
```

Phationmationion:
@rijk, hey!
This is listener which provides duration of media in Vimeo player. It explains observable behavior:
https://github.com/CookPete/react-player/blob/master/src/players/Vimeo.js#L48

At current time ReactPlayer suggests that duration will be requested only after playing. That behavior of ReactPlayer inherited from some external players (I mean Streamable, but may be not only it).
It's problem, but ReactPlayer forced to represent logic of "slowlest" player by each parameter. So, all seeking before player is ready will applies by ReactPlayer only on "play" event instead of "ready", although most of players provides seeking between "ready" and "play".

But it's possible to get duration earlier for Vimeo:
```javascript
this.player.on('loaded', () => {
  this.player.getDuration().then(duration => {
    this.duration = duration
  })
  this.onReady()
})
```
...(it's possible to get duration even earlier)...
...but I'm not sure that propose the best resolving.

CookPete:
Vimeo duration is fetched on load as of version `0.25.1`. Keep in mind that we still do not know the duration of any media _immediately_, hence why the `onDuration` callback exists.

rijk:
That is awesome @CookPete !

PR #251, opened 2017-10-17
tweak Vimeo duration logic

Phationmationion:
Hey!
It's about that issue: https://github.com/CookPete/react-player/issues/250
If we want to add setDuration logic in `this.player.ready().then(...)` instead of `this.player.on("loaded")`, we also should to add the tweak from that PR.

### DIFF
--- a/src/players/Vimeo.js
+++ b/src/players/Vimeo.js
@@ -28,7 +28,9 @@ export default class Vimeo extends Base {
     const id = url.match(MATCH_URL)[3]
     this.duration = null
     if (this.isReady) {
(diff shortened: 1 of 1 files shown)


CookPete:
Good spot. Maybe we should move it to `loaded` event listener instead so we don't have to repeat?

```js
this.player.on('loaded', () => {
  this.onReady()
  this.player.getDuration().then(duration => {
    this.duration = duration
  })
})
```

Phationmationion:
@CookPete https://github.com/CookPete/react-player/issues/250#issuecomment-336959713 :)
Can I do it?

PR #253, opened 2017-10-19
seekOnPlay->seekOnDuration

Phationmationion:
Hi, there!
At current time, when seeking before ActivePlayer is `ready`, ReactPlayer store seeking value and await for `play`. But if we don't playing by default and user seeking when player is not `ready`, we expect seeking at once when it possible. But at current time (in described case) user should press on `play`: before it seeking have not applied.
That PR fixes that behavior.

Inside of PR:
- `seekOnPlay` -> `seekOnDuration`
- protection from Infinity in `getDuration` of Twitch
- removing `this.onDurationCheck()` from `this.onPlay()` of Base (in code from PR `this.props.onDuration()` executes once per media, excluding logic of DailyMotion player)
- adding `this.checkDurationChange()` in Twitch
- passing Infinity to `this.props.onDuration(duration)` in Twitch when playing channel

**!!! IMPORTANT !!!**
DailyMotion resource is not available for my country... It have listener `onDurationChange` and should works good. But it needs for test of emitting `this.props.onDuration()` anyway.

### DIFF
--- a/src/players/Base.js
+++ b/src/players/Base.js
@@ -9,7 +9,7 @@ export default class Base extends Component {
   static defaultProps = defaultProps
   isReady = false
   startOnPlay = true
-  seekOnPlay = null
+  seekOnDuration = null
   componentDidMount () {
     const { url } = this.props
     this.mounted = true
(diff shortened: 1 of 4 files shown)


CookPete:
@Phationmationion The `player-composition` branch is not too far away.. Might be worth opening a PR against that to avoid conflicts later on?

Phationmationion:
@CookPete, 
When I was doing it, I've imagined how it'll seems in composition-structure. And it'll seems better :)
So, I can do it in branch of `player-composition`. Can you give me some callback about my last proposes for composition-structure? I've sended list of proposed changes on your email. I await for your callback to merge my and yours `player-composition` branches.

PR #254, opened 2017-10-20
properly destroy HLS.JS and / or DASH.JS instances

bstst:
I noticed the hls.js instance is still downloading the m3u8 playlist after the player is destroyed, so I figured this should fix this. Also, this way we're properly clearing up memory leaks.

### DIFF
--- a/src/players/FilePlayer.js
+++ b/src/players/FilePlayer.js
@@ -114,7 +114,10 @@ export default class FilePlayer extends Base {
   stop () {
     this.player.removeAttribute('src')
     if (this.hls) {
-      this.hls.detachMedia()
+      this.hls.destroy()
+    }
+    if (this.dash) {
+      this.dash.reset()
     }
   }
   seekTo (amount) {


src/players/FilePlayer.js
```diff
@@ -49,6 +49,12 @@ export default class FilePlayer extends Base {
     }
   }
   componentWillUnmount () {
+    if (this.hls) {
+      this.hls.destroy()
+    }
+    if (this.dash) {
+      this.dash.reset()
```
CookPete: I wonder if this stuff should move to `stop()` instead? `stop()` is called when unmounting, and we are already doing `hls.detachMedia()` there too.

bstst: Oh sorry, didn't notice it there. That's what you get when in a hurry. It makes perfect sense, so I'll "fix the fix".

CookPete:
Thanks! 👍

PR #257, opened 2017-10-22
Loop when player is running

SkinyMonkey:
I need the loop option to be activable even if the player is already playing.

I updated demo accordingly to test if you want to check it.

Thank you for this great component by the way!

### DIFF
--- a/src/demo/App.js
+++ b/src/demo/App.js
@@ -26,7 +26,8 @@ export default class App extends Component {
     played: 0,
     loaded: 0,
     duration: 0,
-    playbackRate: 1.0
+    playbackRate: 1.0,
+    loop: false,
   }
   load = url => {
     this.setState({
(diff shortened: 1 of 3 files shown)


src/demo/App.js
```diff
@@ -132,7 +137,7 @@ export default class App extends Component {
               onPause={this.onPause}
               onBuffer={() => console.log('onBuffer')}
               onSeek={e => console.log('onSeek', e)}
-              onEnded={() => this.setState({ playing: false })}
+              onEnded={() => this.setState({ playing: loop ? true : false })}
```
SkinyMonkey: Useless ternary, 'don't know what I was thinking about.

CookPete:
I like the idea, but don't forget to run [`npm run lint`](https://github.com/CookPete/react-player/blob/master/CONTRIBUTING.md#linting) before commiting.

SkinyMonkey:
Sorry for that, It should be fixed now.

CookPete:
Nice work 👍

PR #289, opened 2017-12-05
Vimeo On Demand URL Support

billdybas:
Vimeo has On Demand URLs like https://vimeo.com/ondemand/tinact/84954874

Currently the player doesn't recognize them, so this PR adds support for them.

I also added some tests for other Vimeo URLs.


CookPete:
Nice one 👍

PR #291, opened 2017-12-11
Prevent a runtime error in FilePlayer

ArsalanDotMe:
I'm constantly getting loads of errors due to `this.player` being null when `getDuration()` is called inside FilePlayer. This commit checks if the player reference exists before using it. Please merge and update soon. Thanks!


ArsalanDotMe:
scratch that. Made changes in the wrong branch. Sending a new PR. How can I offer a patch for the old stable version?

PR #292, opened 2017-12-11
Prevent a runtime error in FilePlayer

ArsalanDotMe:
I'm constantly getting loads of errors due to this.player being null when getDuration() is called inside FilePlayer. This commit checks if the player reference exists before using it. Please merge and update soon. Thanks!


ArsalanDotMe:
To be honest I'm not sure which branch I should have made the PR to but I guess you can guide me when you see this.

CookPete:
Does this error still happen with the `1.0` beta if you `npm install react-player@next`? It's been in the wild for nearly 2 months with no complaints so I will release it pretty soon. I want to avoid branching from `0.25` alongside `1.0` to prevent the auto-generated changelog getting messed up. `1.0` should be stable enough to use.

I'd be interested to see how to reproduce this error, if you have an example repo or jsFiddle.

Also keep in mind that [you don't need to include the `dist` files](https://github.com/CookPete/react-player/blob/master/CONTRIBUTING.md#dist-files) or the changelog in your commits. But I think that might just be because you forked from an older version of the repo.

ArsalanDotMe:
I haven't tested it in the 1.0 beta. Basically these errors get thrown when I'm navigating away from the page that hosts multiple instances of react-player. So something to do with what happens when the component is being unmounted. I'll try to investigate this more and create a jsfiddle when I have an idea of exactly how and when this error gets thrown. But I think it is still good practice to check player for null before accessing its properties

ArsalanDotMe:
Yes, so this happens when the player is in loading state (the video has not started playing yet but the player has requested the media) and the player is unmounted before it gets a chance to download the video.

CookPete:
>But I think it is still good practice to check player for null before accessing its properties

I agree, and that's why I've added a [`callPlayer` util function](https://github.com/CookPete/react-player/blob/c881e37cae78f5d3136b3f8a440dda62ca1c3275/src/utils.js#L95-L109) in `1.0`, but it's probably even better practice to find out _why_ it is trying to call methods on a null player and prevent the root cause.

It would be great to see a jsFiddle or repo where this can be reproduced, as I cannot reproduce it myself from your description. It isn't obvious why it would happen considering:

* [All `FilePlayer` event listeners are removed before unmounting](https://github.com/CookPete/react-player/blob/c881e37cae78f5d3136b3f8a440dda62ca1c3275/src/players/FilePlayer.js#L52)
* [The `onProgress` timeout](https://github.com/CookPete/react-player/blob/c881e37cae78f5d3136b3f8a440dda62ca1c3275/src/ReactPlayer.js#L29) (that calls `getDuration` every second) is cleared before unmounting

>I haven't tested it in the 1.0 beta.

Give it a go. It is stable enough to use and will be published soon anyway.

CookPete:
I'll close this off for now, assuming the problem does not occur in the `1.0.0` beta. If it does, feel free to open a new issue.

PR #293, opened 2017-12-18
Do not load hls.js on iOS

cslansing:
Safari mobile does not support media source extensions and thus will not work with hls.js. Use native HLS video support on iOS instead.

https://caniuse.com/#feat=mediasource


CookPete:
Nice one. I was considering pulling the [`isSupported`](https://github.com/video-dev/hls.js/blob/master/src/helper/is-supported.js) util from the hls.js library but this is probably cleaner and solves most use cases 👍

PR #305, opened 2018-01-15
wip! Refactor player inheritance

JCQuintas:


CookPete:
@JCQuintas not sure what you are trying to achieve here? This would revert the player inheritance logic to an old state.

Issue #322, opened 2018-02-07
Support for SoundCloud Sets


pugson:
SoundCloud has introduced some changes to playlists and they're now called "sets". [An example of one here](https://soundcloud.com/ekalimusic/sets/awakening-mix-series).

The player thinks the URL is a video and tries to render a `<video>` element instead of a SoundCloud iframe. However setting the URL as the one from the API (https://api.soundcloud.com/playlists/333483192) embeds the player correctly.

Would be great to have this working. Thanks for this great library! 💯

PR #324, opened 2018-02-07
Update SoundCloud regex to match Sets

pugson:
Fixes https://github.com/CookPete/react-player/issues/322


CookPete:
In fact, maybe we should change the pattern to just

```regex
/(soundcloud.com|snd.sc)\/.+$/
```

Because we may as well send any soundcloud URL to the widget and let it error if invalid, rather than falling back to trying to play it in a `<video>` element.

pugson:
Yes, good call. Thanks!

PR #332, opened 2018-02-16
Match YouTube URL with extra parameters

WebMaestroFr:
If the regular expression expects the end of string after a video ID, many YouTube links (like https://youtu.be/YE7VzlLtp-4?t=17s for instance) won't match and pass the `canPlay` test.


WebMaestroFr:
`Chrome was not killed in 2000 ms, sending SIGKILL.`, `The command "npm run test" exited with 1.` says Travis...

But I just changed a single character in a regex, what could I've done wrong ?

CookPete:
>Chrome was not killed in 2000 ms, sending SIGKILL., The command "npm run test" exited with 1. says Travis...
>
>But I just changed a single character in a regex, what could I've done wrong ?

Don't worry too much. Testing in Travis has been _extremely_ inconsistent recently, which is very frustrating. I'm trying to work out a more reliable way to run tests, even if it means making them more basic and reducing coverage.

PR #336, opened 2018-02-22
Fixes the onPause event propogation

yashatgit:
If we pass an `onPause` handler to VideoPlayer, the event is not propagated and hence not able to extract relevant info from the event handler.

```
class VideoCaptureView extends Component {
  onVideoPause = event => {
    //can't access event.currentTarget.currentTime
  };
  render() {
    <VideoPlayer url={videoUrl} onPause={this.onVideoPause} />;
  }
}

```

### DIFF
--- a/src/Player.js
+++ b/src/Player.js
@@ -116,10 +116,10 @@ export default class Player extends Component {
     }
     this.onDurationCheck()
   }
-  onPause = () => {
+  onPause = (e) => {
     this.isPlaying = false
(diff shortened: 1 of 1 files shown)


CookPete:
Keep in mind that not all players will pass an event to `onPause` (in fact I think it is only `FilePlayer` that does).

You can get the current time of any player at any time with the [`getCurrentTime` method](https://github.com/CookPete/react-player#instance-methods).

PR #348, opened 2018-03-04
Fix Vimeo link validation.

pawelnvk:
There is an issue, when using player with some of Vimeo links. 

Sharing this video https://vimeo.com/249675407 using Vimeo's share button generates link as follows: https://vimeo.com/yannickcerrutti/annecydroneinmotion. Such link isn't treated as Vimeo's, but it should be. I propose simplifying regex - there is no need for such complexity due to of possible variations of links. 

I've also spotted corrupted file link which I've replaced with working one so that test would pass again.

### DIFF
--- a/src/players/Vimeo.js
+++ b/src/players/Vimeo.js
@@ -5,7 +5,7 @@ import createSinglePlayer from '../singlePlayer'
 
 const SDK_URL = 'https://player.vimeo.com/api/player.js'
 const SDK_GLOBAL = 'Vimeo'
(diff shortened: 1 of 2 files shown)


CookPete:
The tests were failing because your implmentation broke the URL switching logic when the Vimeo player was already loaded. A numeric video ID was required for loading a new video into an existing player.

As of `1.3.0`, all vimeo URLs should now be supported, and switching to another vimeo URL will simply load a new player instance.

pawelnvk:
It's super weird. Locally tests were passing. Error in Travis Ci occurred in different place in one of the test runs.
Anyway, thanks for help and introducing this feature. I really appreciate it.

CookPete:
>It's super weird. Locally tests were passing. Error in Travis Ci occurred in different place in one of the test runs.

Yeah.. https://github.com/CookPete/react-player/pull/332#issuecomment-366624148

PR #355, opened 2018-03-12
added hls error recovery

iambumblehead:
add attempt to recover from hls errors

### DIFF
--- a/src/players/FilePlayer.js
+++ b/src/players/FilePlayer.js
@@ -95,7 +95,7 @@ export class FilePlayer extends Component {
       getSDK(HLS_SDK_URL, HLS_GLOBAL).then(Hls => {
         this.hls = new Hls(this.props.config.file.hlsOptions)
         this.hls.on(Hls.Events.ERROR, (e, data) => {
-          switch(data.type) {
+          switch (data.type) {
             case Hls.ErrorTypes.NETWORK_ERROR:
               // try to recover network error
               // hls: fatal network error encountered, trying to recover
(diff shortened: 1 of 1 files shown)


iambumblehead:
@CookPete it seems travic-ci is unable to complete the test sequence. Would you take a look at my PR?

CookPete:
It seems like a `MEDIA_ERROR` is occuring pretty consistently when seeking towards the end of a HLS video. `recoverMediaError()` is causing a loop to occur as it attempts to play the media immediately and causes another `MEDIA_ERROR`, whereas the demo video seems to recover automatically without having to call `recoverMediaError()`.

I think we are better off passing `Hls.Events.ERROR` through to the `onError` callback, and possibly passing through the `hls` object. Then users can just implement any sort of recovery logic that they want.

CookPete:
As of `1.3.0`, hls.js errors are now passed as `onError(e, data, hlsInstance, hlsGlobal)`

You could implement the same recovery logic like so:

```jsx
function onError (e, data, hlsInstance, hlsGlobal) {
  if (!hlsGlobal) {
    // Probably not a HLS error
    return
  }
  switch (data.type) {
    case hlsGlobal.ErrorTypes.NETWORK_ERROR:
      // try to recover network error
      // hls: fatal network error encountered, trying to recover
      hlsInstance.startLoad()
      break
    case hlsGlobal.ErrorTypes.MEDIA_ERROR:
      // hls: fatal media error encountered, trying to recover
      hlsInstance.recoverMediaError()
      break
    default:
      // cannot recover
      hlsInstance.destroy()
      break
  }
}

<ReactPlayer url={HLS_URL} onError={onError} />
```

Although beware of causing loops when seeking.

iambumblehead:
@CookPete I copied your recommendation here. its a great starting point :), thanks

Issue #362, opened 2018-03-23
Possibility of allowing to add your custom player


BrooklynKing:
I want to know your feels about idea of allowing to pass custom player. If you OK with this idea, i could create PR with couple of approaches. It could look something like
```javascript
import ReactPlayer from 'react-player';
import MyOwnCustomPlayer from './somewher';

ReactPlayer.addPlayer(MyOwnCustomPlayer);
```

AndyBarron:
I would love a feature like this! We use a variety of players that aren't supported by react-player natively, but allowing us to add custom inner players would solve this.

PR #364, opened 2018-03-26
API for adding additonal players to ReactPlayer in runtime

BrooklynKing:
#362

### DIFF
--- a/README.md
+++ b/README.md
@@ -234,6 +234,23 @@ If you aren’t using React, you can still render a player using the standalone 
 
 See [`jsFiddle` example](https://jsfiddle.net/krkcvx9s/)
 
+#### Adding your own players
+
+If you have your own player, that is compatible with ReactPlayer's internal architecture, you can use it like this:
+
+```javascript
+import YourOwnPlayer from './somewhere';
+ReactPlayer.addCustomPlayer(YourOwnPlayer);
+```
+
+Or you can clear all additional players:
+
+```javascript
+ReactPlayer.removeCustomPlayers();
+```
+
+It is your responsibility to ensure that custom players keep up with any internal changes to ReactPlayer in later versions.
+
 #### Using Bower
 
 ```bash
--- a/src/ReactPlayer.js
+++ b/src/ReactPlayer.js
@@ -9,12 +9,20 @@ import renderPreloadPlayers from './preload'
 
 const SUPPORTED_PROPS = Object.keys(propTypes)
 
+let customPlayers = []
+
 export default class ReactPlayer extends Component {
+  static addCustomPlayer = player => {
+    customPlayers.push(player)
+  }
+  static removeCustomPlayers = () => {
+    customPlayers = []
+  }
   static displayName = 'ReactPlayer'
   static propTypes = propTypes
   static defaultProps = defaultProps
   static canPlay = url => {
-    for (let Player of players) {
+    for (let Player of [ ...customPlayers, ...players ]) {
       if (Player.canPlay(url)) {
         return true
       }
@@ -51,7 +59,7 @@ export default class ReactPlayer extends Component {
     this.player.seekTo(fraction)
   }
   getActivePlayer (url) {
-    for (let Player of players) {
+    for (let Player of [ ...customPlayers, ...players ]) {
       if (Player.canPlay(url)) {
         return Player
       }
--- a/test/specs/ReactPlayer.js
+++ b/test/specs/ReactPlayer.js
@@ -4,6 +4,7 @@ import React from 'react'
 import { render, unmountComponentAtNode } from 'react-dom'
 
 import ReactPlayer from '../../src/ReactPlayer'
+import { FilePlayer } from '../../src/players/FilePlayer'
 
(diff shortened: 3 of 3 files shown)


src/ReactPlayer.js
```diff
@@ -56,6 +69,11 @@ export default class ReactPlayer extends Component {
         return Player
       }
     }
+    for (let Player of additionalPlayers) {
```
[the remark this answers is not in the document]
BrooklynKing: @AndyBarron yeah, we can, but i think we should as @CookPete about his opinion.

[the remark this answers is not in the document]
CookPete: I agree that these new loops should go above the regular players.

You could also do:

```js
for (let Player of [ ...additionalPlayers, ...players ]) {
```

for the single loop.

CookPete:
My only thought on this is that passing `additionalPlayers` as a prop might be a bit cleaner and more React-friendly, although we would lose the ability to check the new players in the `canPlay` static.

README.md
```diff
@@ -234,6 +234,23 @@ If you aren’t using React, you can still render a player using the standalone
...
+ReactPlayer.addAdditionalPlayer(YourOwnPlayer);
+```
+
+Or you can clear all additional players:
+
+```javascript
+ReactPlayer.clearAdditionalPlayers();
+```
+
+All responsibilities for keeping your player's API compatible with ReactPlayer is on you.
```
CookPete: Not sure this line is necessary? Seems a bit aggressive.

README.md
```diff
@@ -234,6 +234,23 @@ If you aren’t using React, you can still render a player using the standalone
 
 See [`jsFiddle` example](https://jsfiddle.net/krkcvx9s/)
 
+#### Adding your own players
+
+If you have your own player, that compatible with ReactPlayer's internal architecture, you can use it like this:
```
CookPete: > If you have your own player, that **is** compatible...

README.md
BrooklynKing: @CookPete i just wanted to protect you from issues related to changes in interfaces of your "internal" players.

BrooklynKing:
@CookPete about passing players as props, i'm not sure about it, since with this additional players we kinda make configuration for ReactPlayer as module. If we passing it with props, then we kinda mixing properties of different abstraction layers. And if you passing new props, you expecting changes. And what changes should be when you pass new set of players? Should we run whole process of getting and rendering proper player for passed url? Don't really know what is the best way, so i took simple way with static method and "configuration" approach

README.md
```diff
@@ -234,6 +234,23 @@ If you aren’t using React, you can still render a player using the standalone
...
+
+```javascript
+import YourOwnPlayer from './somewhere';
+ReactPlayer.addAdditionalPlayer(YourOwnPlayer);
+```
+
+Or you can clear all additional players:
+
+```javascript
+ReactPlayer.clearAdditionalPlayers();
```
CookPete: Slightly anal, but I think I prefer referring to the extra players as "custom players" rather than "additional players" throughout the codebase.

BrooklynKing: @CookPete so then `ReactPlayer.addCustomPlayer()` and `ReactPlayer.clearCustomPlayers()`?

CookPete: `addCustomPlayer` and `removeCustomPlayers` would be better I think 👍

BrooklynKing:
@CookPete i updated PR

CookPete:
Nice one! Thanks @BrooklynKing and @AndyBarron.

CookPete:
Published in `1.4.0`. Sorry for the delay!

shawninder:
@BrooklynKing @AndyBarron could either of you share an example of successfully using a custom player with this API?

BrooklynKing:
@shawninder not sure i can share example, since it's in private repo, but i can try to answer on your questions

shawninder:
I'm getting weird errors I can't figure out with my custom player, yet it's basically copy-pasted from the FilePlayer... Here's a question. Did you also use `createSinglePlayer` as react-player uses for its players, or did you start from scratch?

If you're curious, here is my work-in-progress: https://github.com/shawninder/music/blob/master/components/IndexedDBPlayer.js It currently doesn't do anything (I was hoping to stop the Component from crashing the page before starting to add functionality) as it's just the FilePlayer copied, renamed, and with the File stuff removed. I just don't understand why this crashes the page whereas the FilePlayer included in react-player doesn't.

CookPete:
@shawninder are you doing 

```js
import { IndexedDBPlayer } from '...'
```

or 

```js
import IndexedDBPlayer from '...'
```

as the latter would cause issues.

shawninder:
Ah, damn it, I am using `import IndexedDBPlayer from '...'`. Thanks @CookPete, I'm sure this is my problem (will confirm tomorrow). It's times like these when I miss working with a team.

shawninder:
@CookPete While you're in the area, here's a question for you: If I do get this IndexedDBPlayer working, would you be interested in a pull request adding the player to this repo? I'm guessing not, since playing media from IndexedDB feels like an infrequent use-case, but if you do want it, just tell me and I'll submit a pull request when I'm "done".

Issue #463, opened 2018-08-09
Error when removing file based video during onProgress function


zzh8829:
If we remove a video after an onProgress call is queued, a strange null pointer error will happen.
The null error happens at [Player.getCurrentTime](https://github.com/CookPete/react-player/blob/52d98adedb31898448bb7e84c3d674a2d4b8d67d/src/Player.js#L87) in the progress callback. The [FilePlayer.getCurrentTime](https://github.com/CookPete/react-player/blob/52d98adedb31898448bb7e84c3d674a2d4b8d67d/src/players/FilePlayer.js#L165) call might fail because FilePlayer.player is already removed from Dom and there is no null check inside of FilePlayer.

CookPete:
How exactly are you causing this to happen? 

The real solution here would be to [unmount](https://github.com/CookPete/react-player/blob/master/src/Player.js#L25) your player correctly to prevent any further progress callbacks from firing. If you are removing nodes from the DOM outside of the React rendering lifecycle you are asking for errors like this.

CookPete:
Should be fixed in `1.6.5` which includes https://github.com/CookPete/react-player/pull/465

PR #465, opened 2018-08-16
Check null before accessing FilePlayer properties

zzh8829:
This is a fix for https://github.com/CookPete/react-player/issues/463
Adding null check here solves the problem


CookPete:
I'm happy to merge this as a guard for future errors, but note my comment in #463 that this isn't really a solution to the problem if you are removing stuff from the DOM.

Issue #481, opened 2018-09-04
bug, dynamic update of <source(s)> and <track(s)> seems to be ignored 


belgattitude:
Took me a while to figure out, and it's not related to ReactPlayer but good to keep a note. 

(update seems to affect Chrome too: play with effect https://soluble.io/react-player/demo/ (click multiple bunny then multiple elephant))

Setting new *FilePlayer* sources urls or tracks on an already loaded video does not actually reload the   `<video> url` under Firefox (at least Firefox 61 - Ubuntu 18.04). Chrome ~~works as expected~~ does not work neither. 

Setting the video url as a string works, I'm referring specifically to  multiple sources or tracks like:

```
<ReactPlayer
  playing
  url={[
    {src: 'foo.webm', type: 'video/webm'},
    {src: 'foo.ogg', type: 'video/ogg'}
  ]},
 config={{ file: {
    tracks: [
      {kind: 'subtitles', src: 'subs/subtitles.en.vtt', srcLang: 'en', default: true},
      {kind: 'subtitles', src: 'subs/subtitles.de.vtt', srcLang: 'de'}
    ]
  }}}
/>
```

To be sure I does not come from my code, I've updated the demo app.js to include 2 buttons with multiple sources. To reproduce: clone and yarn start https://github.com/belgattitude/react-player.git or see https://github.com/CookPete/react-player/compare/master...belgattitude:test-video-url-change-ignored?expand=1

Changing the source location from Firebug is also affected.

This bug is very annoying in my case, because video are loaded from a list... Props urls are well updated.

belgattitude:
I'm currently trying to figure out how I can circumvent this bug. Any help greatly appreciated.. from MDN (https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement), I've seen we can reload the MediaElement:

> **HTMLMediaElement.load()**
> Resets the media element and restarts the media resource. Any pending events are discarded. How much media data is fetched is still affected by the preload attribute. This method can be useful for releasing resources after any src attribute and source element descendants have been removed. Otherwise, it is usually unnecessary to use this method, unless required to rescan source element children after dynamic changes.

Would it be possible in the library to call such a method on `<source>` or `<tracks>` update ?

belgattitude:
To see the effect in action:

https://soluble.io/react-player/demo/

First click on 'Multiple Bunny' then 'Multiple Elephant' (elephant video loading will be ignored)...  Repeat switching between them and you'll start seing:

```
[...]

PR #482, opened 2018-09-04
Fix: signal multiple sources change to the browser

belgattitude:
Hey @CookPete 

I'm building a PWA where I need to be able to load videos from a list. I'm using multiple sources mode (webm/mp4 and multilingual tracks to set videos).

Unfortunately, if the initial loading works, subsequent updates of the sources does not trigger a reload (although the DOM  shows updated `<source> and <track>`). See my initial ticket : #481. To illustrate the problem switch multiple sources on this demo https://soluble.io/react-player/demo/

I thought initially to change the react keys of the sources in `FilePlayer.js`:

```js
 renderSourceElement = (source, index) => {
    /* This was a first attempt to fix https://github.com/CookPete/react-player/issues/481
     * Unfortunately it's not working
    */
    if (typeof source === 'string') {
      return <source key={`${source}-${index}`} src={source} />
    }
    return <source key={`${source.src}-${index}`} {...source} />
    
```

But it's not working.

My solution was to signal / trigger the change by setting `srcObject = null` when there's a source change. Both Firefox 61 and Chrome 68 seems to work (using ubuntu 18.04). To be sure it does not produce errors on others browsers, it's encloded in a try/catch...

Please let me know if it looks good to you. 

Thanks for the great work.

### DIFF
--- a/src/Player.js
+++ b/src/Player.js
@@ -1,6 +1,7 @@
 import React, { Component } from 'react'
 
 import { propTypes, defaultProps } from './props'
+import { isEqual } from './utils'
 
 const SEEK_ON_PLAY_EXPIRY = 5000
 
@@ -32,7 +33,7 @@ export default class Player extends Component {
   componentWillReceiveProps (nextProps) {
     // Invoke player methods based on incoming props
     const { url, playing, volume, muted, playbackRate } = this.props
-    if (url !== nextProps.url) {
+    if (!isEqual(url, nextProps.url)) {
       if (this.isLoading) {
         console.warn(`ReactPlayer: the attempt to load ${nextProps.url} is being deferred until the player has loaded`)
         this.loadOnReady = nextProps.url
--- a/src/players/FilePlayer.js
+++ b/src/players/FilePlayer.js
@@ -120,14 +120,11 @@ export class FilePlayer extends Component {
     }
 
     if (url instanceof Array) {
-      // Whenever working with multiple sources (array), it seems
-      // to be required to signal manually the browser of the source change.
-      // Just replacing children source dom nodes is not enough
-      try {
-        this.player.load()
-      } catch (e) {
-        // ignore
-      }
+      // When setting new urls (<source>) on an already loaded video,
+      // HTMLMediaElement.load() is needed to reset the media element
+      // and restart the media resource. Just replacing children source
+      // dom nodes is not enough
+      this.player.load()
     } else if (isMediaStream(url)) {
       try {
         this.player.srcObject = url
--- a/src/players/FilePlayer.js
+++ b/src/players/FilePlayer.js
@@ -119,7 +119,7 @@ export class FilePlayer extends Component {
       })
     }
 
-    if (url instanceof Array ) {
+    if (url instanceof Array) {
       // Whenever working with multiple sources (array), it seems
       // to be required to signal manually the browser of the source change.
       // Just replacing children source dom nodes is not enough
--- a/src/players/FilePlayer.js
+++ b/src/players/FilePlayer.js
@@ -124,7 +124,7 @@ export class FilePlayer extends Component {
       // to be required to signal manually the browser of the source change.
       // Just replacing children source dom nodes is not enough
       try {
-        this.player.srcObject = null
+        this.player.load()
       } catch (e) {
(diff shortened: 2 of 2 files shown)


CookPete:
It looks like the "proper" fix for this would be to [update the `key` of the `<video>` to indicate a new element is required](https://stackoverflow.com/a/47382850/1250405), but this breaks the current behaviour for adding/removing event listeners to the DOM nodes – it creates a new DOM node without the event listeners set when the component mounts.

Is it possible that just calling `this.player.load()` would do the trick instead of `srcObject = null`? If so, that seems a bit less hacky and probably wouldn't need a try/catch.

Either way – thanks so much for the amount of effort you've put into this. It's really nice to get such a well-explained issue, complete with a proposed solution.

belgattitude:
> Is it possible that just calling `this.player.load()` would do the trick instead of `srcObject = null`? 

Seems to work as expected with Chrome 69 and Firefox 62, and yes this looks better, so I've updated the P/R with the load() method. Thanks.

> Either way – thanks so much for the amount of effort you've put into this. It's really nice to get such a well-explained issue, complete with a proposed solution.

Ahah... thank you in return ;) 

BTW, you can test with the latest changes here, I guess it won't really be easy to add proper testing at that level:

https://soluble.io/react-player/demo-fixed/

Previous version
https://soluble.io/react-player/demo/

belgattitude:
For information, this P/R does not yet fix issues with subtitles changes (Firefox)... I was unsure of how to do it in the library, so I did it in my code, see the `shouldComponentUpdate()` and `XXXSubtitle()` methods as a example workaround: 

```js
shouldComponentUpdate(nextProps: VideoPlayerProps, nextState: VideoPlayerState): boolean { 
  if (nextProps.video.videoId !== this.props.video.videoId) {
      if (this.playerRef.current !== null) {
          // Let's explicitly load the new video :)
	  const videoEl = this.playerRef.current!.getInternalPlayer() as HTMLVideoElement;
	  // Ohoh Firefox, you keep the old subtitles... let's fix that ;)
	  hideAllSubtitles(videoEl);
	  videoEl.load();
      }
      return true;
  }
  // To be tested, a better solution must be found
  if (nextProps.activeSubtitleLang !== this.props.activeSubtitleLang) {
      const videoEl = this.playerRef.current!.getInternalPlayer() as HTMLVideoElement;
      showSubtitle(videoEl, nextProps.activeSubtitleLang || 'en');
      return true;
  }
  return false;
}
```
```js
const hideAllSubtitles = (video: HTMLVideoElement): void => {
  for (let i = 0; i < video.textTracks.length; i++) {
      console.log('Hidding', video.textTracks[i]);
      video.textTracks[i].mode = 'hidden';
      console.log('New value', video.textTracks[i]);
  }
}
const showSubtitle = (video: HTMLVideoElement, lang: string): void => {
  for (let i = 0; i < video.textTracks.length; i++) {
      if (video.textTracks[i].language === lang) {
	  console.log('SHOWING', video.textTracks[i]);
	  video.textTracks[i].mode = 'showing';
      } else {
	  video.textTracks[i].mode = 'hidden';
      }
  }
}
```

I keep it for information only.... once I have more time to look at it, I'll make a suggestion.

CookPete:
Nice work @belgattitude. I'm assuming that the `key` solution would fix subtitles (and most other problems around this) but we just need to solve the problem with event listeners being lost.

belgattitude:
Hi @CookPete, 

Equality for url's seems to not take arrays into consideration, see the proposal fix:  https://github.com/CookPete/react-player/pull/482/commits/eaaea272e7b12a006dfee420667ecd1b329b0eae. Does it make sense to you ? I've runned into little issue with that. Do you want me to add in a separate P/R or included here (I would love to have it merged before) ?

Secondly: 

> I'm assuming that the key solution would fix subtitles (and most other problems around this) but we just need to solve the problem with event listeners being lost.

I'm not sure to understand this one... Do you mean you'd prefer to  (or we must) implement the `"proper" fix for this would be to update the key of the <video> to indicate a new element is required` way ?  

Thanks for your time !

CookPete:
I'm happy to merge this along with the `isEqual` fix and tackle the alternative fix separately. Thanks for your hard work!

PR #489, opened 2018-09-30
Typings - label in TrackProps

belgattitude:
Added typing for `label` in `TrackProps` -  A user-readable title of the text track which is used by the browser when listing available text tracks.


CookPete:
👌

PR #497, opened 2018-10-12
fix: fix onProgress callback not firing during iOS live streams

ecozoic:
I noticed that the `onProgress` callback only gets fired once during live streams on iOS. This is because, on iOS, `player.duration` returns `Infinity` for live streams. Since it always returns `Infinity`, the progress timeout short-circuits and so the callback stops firing altogether.

I have recently had to fix this issue in a few other places, and I have found that using `player.seekable.end(0)` works great as an alternative. It returns the same thing as `player.duration`, but works on iOS as well.

In order to minimize the potential impact of this change, I explicitly check for `Infinity` and only then do I return `player.seekable.end(0)`. Otherwise I continue to return value of the `duration` property.

Let me know if you need me to me make any updates!

### DIFF
--- a/src/players/FilePlayer.js
+++ b/src/players/FilePlayer.js
@@ -168,7 +168,13 @@ export class FilePlayer extends Component {
   }
   getDuration () {
     if (!this.player) return null
-    return this.player.duration
+    const { duration } = this.player
+    // on iOS, live streams return Infinity for the duration
+    // so instead we use the end of the seekable timerange
+    if (duration === Infinity) {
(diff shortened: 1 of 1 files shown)


ecozoic:
Made some minor updates to improve this fix. While `seekable` should almost always be contiguous anyway, I now make sure to grab the LAST seekable timeRange and no longer make the assumption that it will always be at `0` index.

CookPete:
Nice work! Sorry for the delay on this.

PR #513, opened 2018-11-05
update HLS and DASH SDK to use latest versions

mmaloon:
Use HLS/DASH library url's taken from respective projects' readme's
Prevent loading of HLS/Dash media source if player has already unmounted


CookPete:
There is already [some logic in the `getSDK` util](https://github.com/CookPete/react-player/blob/master/src/utils.js#L64-L66) to prevent loading a resource twice.

I'm wary of always loading the absolute latest version of any SDK. Any future change that may cause an error will break every implementation of ReactPlayer everywhere.

Maybe a better solution would be to hardcode the current latest (after testing it works), then add `hlsVersion` and `dashVersion` to the `config` prop to allow anyone to pass in whatever version they want to use in future?

mmaloon:
de8d961 is actually addressing a different issue from the SDK loading multiple times.  It actually handles the scenario where if you load an HLS file player for the first time, then unmount it before the SDK has async loaded, the SDK load promise resolves and it creates the hlsPlayer and starts loading the HLS src even the react-player is no longer mounted (and hence impossible to access and destroy this newly created hlsPlayer)

PR #544, opened 2018-12-30
playsinline for Android wechat

bingods:


src/players/FilePlayer.js
```diff
@@ -78,6 +78,7 @@ export class FilePlayer extends Component {
     if (playsinline) {
       this.player.setAttribute('playsinline', '')
       this.player.setAttribute('webkit-playsinline', '')
+      this.player.setAttribute('x5-playsinline', 'true')
```
CookPete: Does this need to be `true` or can it be `''` like attributes above?

bingods: '' is ok

CookPete: Ok, let's change it to be consistent and then I'll merge.

```suggestion
      this.player.setAttribute('x5-playsinline', '')
```

CookPete:
Thanks! 👍

Issue #272, opened 2017-11-09
Support for youtube-nocookie.com


Betree:
Youtube provides a service to integrate a video without leaving a cookie if the user does not click on play. It simply use the form `youtube-nocookie.com/embed/youtube_id`.

I'd love to be able to do something like:
```jsx
<ReactPlayer url="https://www.youtube.com/watch?v=dQw4w9WgXcQ" config={{
    youtube: {
        noCookie: true
    }
}}/>
```

That would convert `https://www.youtube.com/watch?v=dQw4w9WgXcQ` to `https://youtube-nocookie.com/embed/dQw4w9WgXcQ` automatically.

However if you think this should be handled by a parent component, I'd like to be at least able to provide a youtube-nocookie url to ReactPlayer (which would be a 14 characters change in the code 😉).

Tell me which one you prefer so I can work on a PR.

Ps: You can check that this is legitimately owned by Google here: https://www.whois.com/whois/youtube-nocookie.com

CookPete:
ReactPlayer uses the [YouTube iFrame API](https://developers.google.com/youtube/iframe_api_reference) to play videos. To support this, the API needs to support it. I can't find any documentation about a `nocookie` mode in the docs or the [custom player params](https://developers.google.com/youtube/player_parameters#Parameters), so I'm not sure this can be done yet.

Adding to the youtube URL pattern wouldn't do much, because it's only used for the `canPlay` method and to parse the 11 character video ID.

Betree:
Thanks for the response, it's indeed undocumented. But you can actually activate the option from Youtube's share menu, so I guess it's not deprecated:

![Screenshot Youtube](https://user-images.githubusercontent.com/1556356/32597455-e53055c4-c58a-11e7-8c67-4746c1c2bf9b.png)

What about implementing it as a new provider that would put the URL directly in a IFrame ?

```
<iframe width="560" height="315" src="https://www.youtube-nocookie.com/embed/dQw4w9WgXcQ" frameborder="0" allowfullscreen></iframe>
```

CookPete:
Yeah I don't think it's deprecated, but it's just not part of the iFrame API.

Unfortunately, it looks like you lose all events and interactions with the player in nocookie mode:

* https://groups.google.com/forum/#!topic/youtube-api-gdata/lscdyDrV0T0
* https://stackoverflow.com/questions/36480181/youtube-player-api-onready-not-firing-for-www-youtube-nocookie-com
[...]

PR #557, opened 2019-01-24
Added host-option to embed youtube from 'nocookie'-domain

AlexandreKilian:
Fix for https://github.com/CookPete/react-player/issues/272 to allow changing the iFrame Domain to "nocookie".
Linted, tested, built and demo updated.


Betree:
This is awesome @AlexandreKilian! Thanks for your work on this. Will definitely be useful for us.

CookPete:
I really didn't want to add a new top level `host` prop for a Youtube-specific edge case. 

In `1.9.0` I've added `embedOptions` to the youtube config that gets spread [here](https://github.com/CookPete/react-player/blob/6d6230a183a5019cbe42d7071f5e194c1a5c2df9/src/players/YouTube.js#L66) after the other options. Then you can do:

```jsx
<ReactPlayer
  url='...'
  config={{ 
    youtube: { 
      embedOptions: { 
        host: 'https://www.youtube-nocookie.com' 
      } 
    } 
  }}
/>
```

Any override any other options in future, if you want to.

PR #568, opened 2019-02-12
Add "playsinline" attribute to avoid iOS automatic full-screen

Maxou44:
By default IOS play videos in full-screen mode, this commit disable this feature to improve consistency between browsers.

More information here: https://webkit.org/blog/6784/new-video-policies-for-ios/

### DIFF
--- a/src/players/FilePlayer.js
+++ b/src/players/FilePlayer.js
@@ -262,6 +262,7 @@ export class FilePlayer extends Component {
         controls={controls}
         muted={muted}
         loop={loop}
+        playsinline=""
         {...config.file.attributes}>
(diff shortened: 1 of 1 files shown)


CookPete:
There is already a `playsinline` prop that does exactly this and [sets various attributes for cross browser compatibility](https://github.com/CookPete/react-player/blob/37700cc6a8c7a1d6c2d1796862ff4aafb45739f9/src/players/FilePlayer.js#L78-L82):

```jsx
<ReactPlayer url={...} playsinline />
```

Maxou44:
Sorry, my bad

CookPete:
No probs! 👍

PR #597, opened 2019-03-15
Do not use vimeo API to play vimeo hosted video files.

rin:
If you have a vimeo PRO account, you get direct urls for your videos. Unfortunately, react-player will try to play them with the vimeo player.

This was fixed for `mp4` files in this commit:
https://github.com/CookPete/react-player/commit/6f10f61a0657ae800b4d2116787efce2ce97f2dd

The issue is still there for `m3u` files though – e.g. 
`https://player.vimeo.com/external/268931179.m3u8?s=4d7bec5817b90f9227891dd828e32deb91fa01e7`.

### DIFF
--- a/src/players/Vimeo.js
+++ b/src/players/Vimeo.js
@@ -6,7 +6,7 @@ import createSinglePlayer from '../singlePlayer'
 const SDK_URL = 'https://player.vimeo.com/api/player.js'
 const SDK_GLOBAL = 'Vimeo'
 const MATCH_URL = /vimeo\.com\/.+/
-const MATCH_FILE_URL = /vimeo\.com\/external\/.+\.[mp4|m3u]/
+const MATCH_FILE_URL = /vimeo\.com\/external\/[0-9]+\..+/
 
 export class Vimeo extends Component {
   static displayName = 'Vimeo'
--- a/src/players/Vimeo.js
+++ b/src/players/Vimeo.js
@@ -6,7 +6,7 @@ import createSinglePlayer from '../singlePlayer'
 const SDK_URL = 'https://player.vimeo.com/api/player.js'
 const SDK_GLOBAL = 'Vimeo'
 const MATCH_URL = /vimeo\.com\/.+/
-const MATCH_FILE_URL = /vimeo\.com\/external\/.+\.mp4/
+const MATCH_FILE_URL = /vimeo\.com\/external\/.+\.[mp4|m3u]/
 
 export class Vimeo extends Component {
   static displayName = 'Vimeo'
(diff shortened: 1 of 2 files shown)


src/players/Vimeo.js
```diff
@@ -6,7 +6,7 @@ import createSinglePlayer from '../singlePlayer'
 const SDK_URL = 'https://player.vimeo.com/api/player.js'
 const SDK_GLOBAL = 'Vimeo'
 const MATCH_URL = /vimeo\.com\/.+/
-const MATCH_FILE_URL = /vimeo\.com\/external\/.+\.mp4/
+const MATCH_FILE_URL = /vimeo\.com\/external\/.+\.[mp4|m3u]/
```
CookPete: Why don't we just look for any extension?
```suggestion
const MATCH_FILE_URL = /vimeo\.com\/external\/[0-9]+\..+/
```

thunderkid:
Hi. I'm glad I just saw this. I was about to spend some unpleasant time figuring out some hacks to make the vimeo player behave more consistently. I didn't realize vimeo files could actually be accessed with a non-vimeo player instead. 

Is there an example of this we could add to the demo? (I'd be happy to provide a PR if I just had a link.) Doing a quick search I couldn't even find a vimeo pro file I could use as a test case but I might just not know what I'm supposed to be looking for. Thanks.

CookPete:
I don't have Vimeo Pro, but if anyone wants to contribute an example of a direct URL, they are more than welcome!

rin:
@thunderkid I don't really think it needs to be added to the demo, because in the end it's just a video file that happens to be hosted on vimeo. The url I used in my test is the only one I have right now (don't have access to the vimeo Pro account myself) and when you change the regex of the vimeo player, it works in the demo, too.

src/players/Vimeo.js
rin: Will try to get access to the vimeo Pro account to find out what these urls look like – but you're probably right, there are more extensions than only `mp4` and `m3u`.

schuylr:
@CookPete The `m3u8` extension is the most common problem right now. Pro links are basically various qualities of MP4 videos and HLS formats.

Here are all of the video formats that we got from the Direct URLs page on our Vimeo Pro account. I think the quality of the MP4 formats are determined by the `profile_id` param:

```
https://player.vimeo.com/external/231900993.hd.mp4?s=<redacted>&profile_id=175

https://player.vimeo.com/external/231900993.hd.mp4?s=<redacted>&profile_id=174

https://player.vimeo.com/external/231900993.sd.mp4?s=<redacted>&profile_id=165

https://player.vimeo.com/external/231900993.sd.mp4?s=<redacted>&profile_id=164

https://player.vimeo.com/external/231900993.m3u8?s=<redacted>
```

PR #612, opened 2019-04-03
FilePlayer: calls onPlay when "playing" events gets fired

me-andre:
According to the README onPlay gets called when media starts or
resumes playing after pausing or buffering.
The corresponding DOM event for that is "playing", not "play", see
https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/playing_event
The "play" event indicates a change of "paused" or "autoplay" attr.

### DIFF
--- a/src/players/FilePlayer.js
+++ b/src/players/FilePlayer.js
@@ -68,7 +68,7 @@ export class FilePlayer extends Component {
   addListeners () {
     const { onReady, onPlay, onPause, onEnded, onError, playsinline, onEnablePIP } = this.props
     this.player.addEventListener('canplay', onReady)
-    this.player.addEventListener('play', onPlay)
+    this.player.addEventListener('playing', onPlay)
     this.player.addEventListener('pause', onPause)
     this.player.addEventListener('seeked', this.onSeek)
     this.player.addEventListener('ended', onEnded)
@@ -84,7 +84,7 @@ export class FilePlayer extends Component {
   removeListeners () {
     const { onReady, onPlay, onPause, onEnded, onError, onEnablePIP } = this.props
     this.player.removeEventListener('canplay', onReady)
-    this.player.removeEventListener('play', onPlay)
+    this.player.removeEventListener('playing', onPlay)
     this.player.removeEventListener('pause', onPause)
     this.player.removeEventListener('seeked', this.onSeek)
     this.player.removeEventListener('ended', onEnded)


CookPete:
>The playing event is fired when playback is **ready to start** after having been paused or delayed due to lack of data.

We want `onPlay` to trigger when a video is played, not when it is _ready to start_.

The exact behaviour for `onPlay` is different for different player APIs depending on how they work. I'm happy to update the readme to reflect this but I don't see any value in changing a core behaviour of the player (that no one has found a problem with until now).

There is always `config.file.attributes` and the `getInternalPlayer()` instance method for you to set whatever attributes or events you like on the `<video>` element.

me-andre:
Wait, you say no one has found a problem, but what about me? :) If I fire an issue, would that be considered a found problem?
The "playing" event is actually gets called when the video starts to play after being paused or staled due to buffering (maybe they used a slightly confusing wording at MDN). I thought THIS is what we want - to call `onPlay` when picture starts moving, you know. This is also what non-file players do, so it's surprises me a bit that you call an inconsistent behavior an expected one.

CookPete:
>Wait, you say no one has found a problem, but what about me? :) If I fire an issue, would that be considered a found problem?

Sorry, I didn't mean to sound dismissive. My point is that the library has been around for nearly 4 years, and no one _until now_ has claimed that the `onPlay` behaviour is wrong.

>The "playing" event is actually called when the video starts to play after being paused or staled due to buffering (maybe they used a slightly confusing wording at MDN). I thought THIS is what we want - to call onPlay when picture starts moving, you know. 

I found a more useful description [here](https://developer.mozilla.org/en-US/docs/Web/Guide/Events/Media_events):

>`playing` | Sent when the media has enough data to start playing, after the play event, but also when recovering from being stalled, when looping media restarts, and after seeked, if it was playing before seeking.

It seems too trigger happy.. I agree that `onPlay` after buffering is useful – no dispute there. But, we don't want `onPlay` to fire when a video loops, because that doesn't make sense. We also don't want `onPlay` to fire after seeking, because that doesn't make sense (and we already have `onSeek`).

>This is also what non-file players do, so it's surprises me a bit that you call an inconsistent behavior an expected one.

I didn't claim that anything was expected behaviour. The point is that `onPlay` should be as consistent as possible across the [number of players](https://github.com/CookPete/react-player/tree/master/src/players) that we have. The event should fire _when the media is played_ and `playing` is fired much more often than that, to the point where the inconsistency between players will cause confusion. eg _"Wait, why did my `onPlay` handler get fired because I seeked/looped the video? It wasn't paused so why would it be called? The Youtube video didn't do that.."_

Also see [this answer on SO](https://stackoverflow.com/a/14270041/1250405):

>So, for practical purposes, `play` is good for setting a visual indicator of play state, like the icon on a play/pause toggle button. Use `playing` if you want to know when the video is actually playing. For example, you might want to show a spinning "loading" icon when the waiting event fires, and then remove that icon when playing fires.

I imagine the vast majority of people using this library are using `onPlay` to update a play/pause toggle.

If you like, I'm happy to add `onPlaying` and `onWaiting` callbacks, but not add them to the readme to avoid confusion (and because they only apply to one player).

me-andre:
Hmmm... one thing that confuses me is why someone ever wants to update "playing" status by receiving `onPlay` while he controls the player by setting `playing=[true|false].`

-----------
The other thing is:
1. YouTube status 1 means the video is playing, as in DOM event "playing", i.e. the picture is moving; you won't get `status=1` when it's buffering. So on each transition between "buffering" and "playing" would result in `onPlay()`.
2. But e.g. when using FB player, we get `"startedPlaying"` when the video starts to play, but when it stops for buffering we receive `"startedBuffering"`. Then if we want to do smth when it resumes, we subscribe to `"finishedBuffering"`.
So with FB we get `onPlay` once, but on YouTube - multiple times. You say that if `onPlay` is called too often, that may be undesired for some people. So, is YouTube player is broken?

-----------
Now I think I should have started with that in form of an issue.
The reason I made this PR is: there's no reliable cb for indicating the video started/finished buffering. So my first decision was to make FilePlayer behave like YouTube (that seemed most "natural" behavior for me).
Now I think there are 2 problems to be addressed:
1. YouTube player fires `onPlay` multiple times (doesn't affect my app, but it may be good to make the behavior consistent across players).
2. There's no cb that gets called when buffering ends (FilePlayer and Facebook are affected, others I haven't checked; YouTube does fire `onPlay`).

I can add `onBuffer` to `FilePlayer` and implement `onFinishedBuffering` (`FilePlayer`, `YouTube` and `Facebook`). What do you think?

me-andre:
@CookPete , please take a look https://github.com/CookPete/react-player/pull/615

Issue #74, opened 2016-06-17
Need callback for 'loading/buffering' state.


andschdk:
When internet connection is bad playback can stop before video is totally loaded. Here no callback is supplied.

This can be useful for showing a spinner when playback stops to buffer more.

I find no way to achieve this from current callbacks.

CookPete:
`onBuffer` will be called when a video has to stop to load before resuming. When it resumes, `onPlay` will be called.

If you are looking to show a spinner when first loading a video, just set `showSpinner: true` or whatever in your initial state and then use the `onStart` callback to set `showSpinner: false`

andschdk:
@CookPete is this also the case if streaming from an URL?
I'm only able to see the behaviour you describe when using youtube/vimeo...

andschdk:
I found these events from `<video>` to be useful.
https://github.com/joblicate/react-player/commit/c42318a085c7c7584b6e9351c759d1654e68df3d

I just need to make a nice way to observe when to display a loading indicator and when to remove it again.

Maybe a callback `onLoadingChange(loading: bool)` where `loading` bool will be true after either a `onwaiting` or `onstalled` event.  `onLoadingChange(loading: bool)` could then report false after `onplaying` event.

Don't know if this makes sense, will look further into it..

CookPete:
Nice spot with `onwaiting` and `onstalled`. I'm happy to accept a PR with those changes (although the `e` param is unused in those listeners).

I think `onBuffer` and `onPlay` is enough to get you the behaviour you want. Just hold `showSpinner: true` (or something) in your state, then:

```js
<ReactPlayer
  url={url}
  playing
  onPlay={() => this.setState({ showSpinner: false })}
  onBuffer=(() => this.setState({ showSpinner: true })}
/>
```

andschdk:
Unfortunately this is very bad supported :-(

`onwaiting` and `onstalled` can be used to know when video needs to buffer the time from playback stops till the event is fired differs from browser to browser. Found one SO post where one claims that the events is not fired in some cases http://stackoverflow.com/questions/21399872/how-to-detect-whether-html5-video-has-paused-for-buffering.

I rely on `onwaiting` and `installed` however, but will consider rewriting this in future. Detecting buffer state based on played progress value can be error prune I think.

When enough data is loaded `onplaying` event should be fired according to the docs at least.. https://developer.mozilla.org/en-US/docs/Web/Events/playing
[...]

PR #613, opened 2019-04-03
FilePlayer calls onBuffer when "waiting" is fired

me-andre:
Introduces support of `onBuffer` callback to the `FilePlayer`

### DIFF
--- a/src/players/FilePlayer.js
+++ b/src/players/FilePlayer.js
@@ -66,9 +66,10 @@ export class FilePlayer extends Component {
     this.removeListeners()
   }
   addListeners () {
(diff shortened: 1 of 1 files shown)


CookPete:
This has already been discussed before: https://github.com/CookPete/react-player/issues/74. The `waiting` event seems too buggy to rely on for `onBuffer`, especially if `onPlay` is then not called when playing is resumed.

There is a `getInternalPlayer()` instance method if you want to set this in your own implementation.

me-andre:
I definitely agree.
Not firing `onPlay` after "waiting" can badly break things, as it did to my app.
This is what is fixed by https://github.com/CookPete/react-player/pull/612

me-andre:
Closing in favor of https://github.com/CookPete/react-player/pull/615

Issue #653, opened 2019-05-26
Player doesn't support dashOptions to add extra methods form dash.js


RajnishKatharotiya:
#### Current Behavior
Well, Dash.js is becoming more and more popular these days we need to enhance some functionality to explore dash through this player.
Currently in filePlayer does support to add custom methods for hlsOptions, but for a dash player, there isn't have any available props to access another method of dash.js. 
The issue was I needed to send cookies while requesting mpd file but the player isn't have enabled a required method for the dash.

#### Expected Behavior
So for access to this method, we can add dashOptions as hlsOptions, something like this.
```jsx
<ReactPlayer
  config={
    file: {
        dashOptions: {
            getXHRWithCredentialsForType: {
                params: ['GET'],
                callback: (e) => console.log(e)
            }
        }
    }
  }
/>
```
 

#### Steps to Reproduce
- This issue can be resolved with a just small change in FilePlayer.js 
- Need to update method for DashPlayer. After initialization of player, we can add a function which will call given function with given params(optional) and pass the response back to react-player through callback(optional) prop.

RajnishKatharotiya:
Added this request to solve this issue.

https://github.com/CookPete/react-player/pull/654

check this code out may this will be helpful.

CookPete:
I think #654 is a quite convoluted solution for this. You can access the dash instance with `getInternalPlayer('dash')` which makes it quite easy to call your own methods:

```jsx
class Player extends Component {
  ref = player => {
    this.player = player
  }
  onReady = () => {
    const dash = this.player.getInternalPlayer('dash')
    console.log(dash.getXHRWithCredentialsForType('GET'))
  }
  render () {
    return (
      <ReactPlayer
        ref={this.ref}
        onReady={this.onReady}
        url='http://dash.edgesuite.net/envivio/EnvivioDash3/manifest.mpd'
        playing
      />
    )
  } 
}
```

eliataylor:
@CookPete, this would not allow setting an `Authorization` header to the XHR object. any ideas for handling that via Dash?

Frikster:
Given there explicitly is an [updateHeaders](https://github.com/Dash-Industry-Forum/dash.js/blob/7a397c9f3b2f02df980104537a4b30200f6dd427/src/streaming/protection/controllers/ProtectionController.js#L719) function in Dash's `ProtectionController` I would have thought either of the two lines in my onError below would do something to fix the issue and add the Auth token:

```js
 class Player extends Component {
[...]

PR #654, opened 2019-05-26
Feature added option for dash

RajnishKatharotiya:
#### Current Behavior
Well, Dash.js is becoming more and more popular these days we need to enhance some functionality to explore dash through this player.
Currently in filePlayer does support to add custom methods for hlsOptions, but for a dash player, there isn't have any available props to access another method of dash.js. 
The issue was I needed to send cookies while requesting mpd file but the player isn't have enabled a required method for the dash.

#### Expected Behavior
So for access to this method, we can add dashOptions as hlsOptions, something like this.
```jsx
<ReactPlayer
  config={
    file: {
        dashOptions: {
            getXHRWithCredentialsForType: {
                params: ['GET'],
                callback: (e) => console.log(e)
            }
        }
    }
  }
/>
```
 

#### Steps to Reproduce
- This issue can be resolved with a just small change in FilePlayer.js 
- Need to update method for DashPlayer. After initialization of player, we can add a function which will call given function with given params(optional) and pass the response back to react-player through callback(optional) prop.


CookPete:
See https://github.com/CookPete/react-player/issues/653#issuecomment-499241450

PR #655, opened 2019-05-28
[d.ts] Fix typing of arguments of onError handler

kuy:
onError handler has 4 arguments: https://github.com/CookPete/react-player/blob/master/src/players/FilePlayer.js#L130

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -102,7 +102,7 @@ export interface ReactPlayerProps {
   onEnded?(): void;
   onEnablePIP?(): void;
   onDisablePIP?(): void;
-  onError?(error: any): void;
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -102,7 +102,7 @@ export interface ReactPlayerProps {
   onEnded?(): void;
   onEnablePIP?(): void;
   onDisablePIP?(): void;
-  onError?(error: any): void;
+  onError?(error: any, data: any, hls: any, cls: any): void;
```
CookPete:
```suggestion
  onError?(error: any, data?: any, hlsInstance?: any, hlsGlobal?: any): void;
```

PR #673, opened 2019-07-05
Vimeo wrapper div should not have a background color.

graemeworthy:
Sometimes a responsive video will have 1px on the side or bottom where this black poked through.


blaineo:
Is there any reason that this can't be merged in? The hard-coded background color is extremely limiting.

PR #714, opened 2019-09-10
change( Removing unconfigurable background color )

blaineo:
Removing the hard coded background color. The user can set a background color to black in they style-able container.


blaineo:
Closed because it is a duplicate.

PR #716, opened 2019-09-22
Add PIP support for Safari

destruc7i0n:
Safari supports PIP and has some benefits over that in Chrome for macOS (ability to use over multiple fullscreen windows, etc.). This PR makes the changes required to support Safari PIP mode.

**Note**
In the current Safari release (13), the `onDisablePIP` is run twice when exiting PIP. This is a bug with Safari has been fixed in the [latest Tech preview](https://webkit.org/blog/9526/release-notes-for-safari-technology-preview-91/).


CookPete:
Nice work! 👍

PR #718, opened 2019-09-23
Properly delete SDK requests

bastei:
The changes of #711 don't have an effect as `requests[url]` is set to an empty array after a failed fetch which evaluates to `true` for all following fetches. It has to be set to something evaluating to `false` instead. I added a test for the scenario.

Additionally, I activated the other getSDK tests. As getSDK() uses global variables, the tests have to be serial.


CookPete:
Nice work 👍 I think I was trying to be too clever when trying to avoid the use of `delete` in the original PR. Apologies!

PR #745, opened 2019-11-13
Support to set the custom facebook version

HsuTing:
The new version of the facebook version is `v3.3`. I would like to have a new config to set the facebook version. In my website, `react-player` will overwrite my facebook version. As a result, I can not use facebook login in my website.

### DIFF
--- a/src/props.js
+++ b/src/props.js
@@ -34,7 +34,8 @@ export const propTypes = {
       preload: bool
     }),
     facebook: shape({
-      appId: string
+      appId: string,
+      version: string
     }),
     dailymotion: shape({
       params: object,
@@ -120,6 +121,7 @@ export const defaultProps = {
     },
     facebook: {
       appId: '1309697205772819'
+      version: 'v2.5'
     },
     dailymotion: {
       params: {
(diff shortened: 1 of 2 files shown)


src/props.js
```diff
@@ -120,6 +121,7 @@ export const defaultProps = {
     },
     facebook: {
       appId: '1309697205772819'
+      version: 'v2.5'
```
CookPete: Maybe we should bump this to `v3.3` if that's the latest? Does everything work as normal?

```suggestion
      version: 'v3.3'
```

HsuTing: I had changed to `v3.3`, and I did not find any problem.

HsuTing:
Hi @CookPete . Does this `pr` have any problem?
Can you merge this `pr` and release?

CookPete:
Published in `react-player@1.14.2`. Thanks @HsuTing !

HsuTing:
Thank you! 👍

HsuTing:
> load plans not working

Can you give me the example or a reproduced repo ?
Maybe I can help you.

PR #758, opened 2019-11-28
update hls.js version

superphung:
By updating hls version, i manage to get the programDateTime from hls instance, otherwise it is undefined


ouadie-lahdioui:
+1

abardan2:
+1

PR #794, opened 2020-02-02
Move @ava/babel to devDependencies

lloydh:
I have encountered build issues resulting from ``@ava/babel``s very strict node ``engines`` config. As it relates to testing it should be fine in devDependencies.


CookPete:
🤦‍♂ Sorry!

CookPete:
Fixed in `react-player@1.15.2`

PR #839, opened 2020-03-26
add Spotify play button

crowd-studio:
While Spotify doesn't offer any solutions, it may be interesting to have this functionality that only shows the play button widget


CookPete:
I appreciate the effort, but this seems fairly redundant? Aren't you better off just rendering an `<iframe>` in your app, compared to loading the entire ReactPlayer library to do the same thing?

crowd-studio:
I think no because in this way you have centralized all the embeds on the same vendor and you only need to send the URL from the database without any extra logic.

PR #851, opened 2020-04-14
fixes: subtitles always one render behind

beatthat:
Subtitles were lagging behind the playing video by one render.
The cause of the bug was that file-member property ReactPlayer::config
would get updated in componentDidUpdate (which happens after render).
The fix was to move the config update step into render


CookPete:
This should be fixed in `v2.0.0`

PR #864, opened 2020-04-27
Adds the missing default data for the selected player

Domino987:
This adds the default config to the config file, if a certain player is selected. 
If not and an empty config is provided, calls like config.tracks or config.attribuites fails.
Also the cdn loading fails, because the version is undefined because the default version is not supplied.


CookPete:
Published in `v2.0.1`. Thanks for the PR!

PR #866, opened 2020-04-27
Fix ReactPlayer refs not being set

alexivanov:
This fixes an issue and where `player` and `wrapper` refs are not set correctly in `ReactPlayer`, causing instance methods to stop working. Tested with v2.0.0 and I think this version is the only one affected.

Introduced in 9c9a4d401f8238ddef5630e0b3727604d84e21ca.

### DIFF
--- a/src/ReactPlayer.js
+++ b/src/ReactPlayer.js
@@ -42,7 +42,7 @@ export default class ReactPlayer extends Component {
     showPreview: !!this.props.light
   }
 
-  refs = {
+  elementRefs = {
     wrapper: wrapper => { this.wrapper = wrapper },
     player: player => { this.player = player }
   }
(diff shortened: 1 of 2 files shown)


src/ReactPlayer.js
```diff
@@ -42,7 +42,7 @@ export default class ReactPlayer extends Component {
     showPreview: !!this.props.light
   }
 
-  refs = {
+  elementRefs = {
```
alexivanov, line 45: `refs` overshadows legacy [`React.Component`](https://reactjs.org/docs/refs-and-the-dom.html#legacy-api-string-refs) prop and is overwritten.

CookPete:
Fixed separately in `v2.0.1`. Cheers for the PR!

Issue #867, opened 2020-04-28
Seeking behavior doesn't seem to work on demo page.


ralusek:
Be sure to [search for your issue](https://github.com/CookPete/react-player/issues) before opening a new one.

#### Current Behavior
On this demo page:
https://cookpete.com/react-player/
The seek position slider will update correctly with the current timestamp, but clicking on a position along the slider will temporarily move the slider to that position, fail to seek the video accordingly, and jump back to the position corresponding with the player's position. I tried this with YouTube, Vimeo, and a few others, and they all demonstrated the same behavior.

#### Expected Behavior
I expect seeking to a position to have the corresponding video playback also seek to the same position.

#### Steps to Reproduce
1.) go to https://cookpete.com/react-player/
2.) click on Youtube Test A
3.) click play on the video, wait for playback to begin
4.) note position on bar titled "Seek" updating correctly, corresponding with position in playback
5.) attempt to move the seek marker (not in the YouTube video, but react-player UI) to a desired position
6.) note that the video does not seek to this position, and the playback marker jumps back to the video playback position

#### Environment
- URL attempting to play: https://www.youtube.com/watch?v=oUFJJNQGwhk
- Browser: Chrome
- Operating system: Mac OS
- Demo page: https://cookpete.com/react-player/

ralusek:
@CookPete I fixed this issue here:
https://github.com/CookPete/react-player/pull/868

All tests/lint passing, I recommend merging it because a lot of critical behavior is resolved by this change.

CookPete:
Fixed in `v2.0.1`

omerkhan8:
working fine with youtube, but this is still happening in vimeo.. any help?

cagarweyne:
> Be sure to [search for your issue](https://github.com/CookPete/react-player/issues) before opening a new one.
> 
> #### Current Behavior
> On this demo page:
> https://cookpete.com/react-player/
> The seek position slider will update correctly with the current timestamp, but clicking on a position along the slider will temporarily move the slider to that position, fail to seek the video accordingly, and jump back to the position corresponding with the player's position. I tried this with YouTube, Vimeo, and a few others, and they all demonstrated the same behavior.
> 
> #### Expected Behavior
> I expect seeking to a position to have the corresponding video playback also seek to the same position.
> 
> #### Steps to Reproduce
> 1.) go to https://cookpete.com/react-player/
> 2.) click on Youtube Test A
> 3.) click play on the video, wait for playback to begin
[...]

PR #868, opened 2020-04-28
Fix player refs not setting correctly.

ralusek:
I described an issue I encountered here on the demo.
https://github.com/CookPete/react-player/issues/867

I then experienced this same problem in my own code, where `seekTo` wasn't working correctly. It turns out that the issue was actually far more widespread, and the ref just wasn't being set at all.

When I logged out the object in `ReactPlayer` that was called `refs`, which was meant to hold two functions to set refs in child components, it was an empty object. This appears to be a reserved keyword that is presumably used by react, as it was being overridden with an empty object. Simply changing the property name from `refs` to `setRef` resolved the issue, and seeking behavior (among other things) now work correctly.

In order to test: follow the steps in the issue I linked above, and notice the broken behavior. Then test the same behavior in this branch, and note it functioning correctly.

### DIFF
--- a/src/ReactPlayer.js
+++ b/src/ReactPlayer.js
@@ -42,7 +42,7 @@ export default class ReactPlayer extends Component {
     showPreview: !!this.props.light
   }
 
-  refs = {
+  setRef = {
     wrapper: wrapper => { this.wrapper = wrapper },
     player: player => { this.player = player }
   }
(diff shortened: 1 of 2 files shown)


CookPete:
There was already a PR for this at https://github.com/CookPete/react-player/pull/866. Regardless, this should be fixed in `v2.0.1`

ralusek:
Oh, nice. Thanks

PR #877, opened 2020-05-02
Match YouTube User Uploads

faizanbhat:
Added ability to set URL to YouTube user profiles e.g. https://youtube.com/user/songstowearpantsto/
e.g. https://www.youtube.com/user/Rhaptapsody

If this URL pattern is matched, the player loads user's uploads as specified in the YouTube Player API. The playlist_type var is set to 'user_uploads' and the list is set to the username.


CookPete:
Nice work 👍

PR #894, opened 2020-05-21
Destroy previous vidyard player before creating new one

schipiga:
I had met a bug, that `react-player` calls several time `Vidyard.api.renderPlayer` if i switch video, which leads several `vidyard` players presence in same page.
Fix is to destroy previous player before creating new one.

### DIFF
--- a/src/players/Vidyard.js
+++ b/src/players/Vidyard.js
@@ -53,7 +53,6 @@ export default class Vidyard extends Component {
 
   stop () {
     window.VidyardV4.api.destroyPlayer(this.player)
-    delete this.player
   }
 
   seekTo (amount) {
--- a/src/players/Vidyard.js
+++ b/src/players/Vidyard.js
@@ -15,9 +15,12 @@ export default class Vidyard extends Component {
     this.props.onMount && this.props.onMount(this)
   }
 
-  load (url, isReady) {
+  load (url) {
     const { config, onError, onDuration } = this.props
     const id = url && url.match(MATCH_URL_VIDYARD)[1]
+    if (this.player) {
+      this.stop()
+    }
     getSDK(SDK_URL, SDK_GLOBAL, SDK_GLOBAL_READY).then(Vidyard => {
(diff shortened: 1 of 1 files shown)


schipiga:
@CookPete could you pls review this pr, it's quite important for in my work

src/players/Vidyard.js
```diff
@@ -19,6 +19,10 @@ export default class Vidyard extends Component {
     const { config, onError, onDuration } = this.props
     const id = url && url.match(MATCH_URL_VIDYARD)[1]
     getSDK(SDK_URL, SDK_GLOBAL, SDK_GLOBAL_READY).then(Vidyard => {
+      if (this.player) {
```
schipiga, line 22: I put this chunk before `if (!this.container) return` because it gives more explicit behaviour. For example:
- User watches video and he clicks link to open another video.
- If container check will be before this chunk and container is missing (somehow), then new video isn't opened and previous video will continue to play. Nothing happens for user.
- If destroy happens before and no container, then previous video will be stop and new video isn't opened. And for user it's more explicit that problem with video playing, but not with link to open video.

src/players/Vidyard.js
```diff
@@ -50,6 +53,7 @@ export default class Vidyard extends Component {
 
   stop () {
     window.VidyardV4.api.destroyPlayer(this.player)
+    delete this.player
```
CookPete, line 56: Is this really necessary?

schipiga: @CookPete I think it's logical. If you destroy object and `this.player` doesn't exist, you don't need to keep reference to non-existing object.

CookPete: I’ve merged this but left this line out for the time being. I’m not a fan of the delete statement and would rather be consistent with the rest of the codebase (which avoids calling delete anywhere)

schipiga: ok clear :) any case, I checked new release with fix and works well

Issue #623, opened 2019-04-11
Wistia onPlay not firing


brandon-rentpingmedia:
#### Current Behavior
If you have the same wistia video on multiple pages within your react application, only the first one will fire the onPlay (and other) events.

#### Expected Behavior
Each wistia video should fire events.

#### Steps to Reproduce
1. Embed a wistia video on a page with onPlay.
2. Embed the same wistia video on another page/popup with onPlay.

#### Environment
- URL attempting to play: Any
- Browser: Any
- Operating system: Any

#### Other Information
The issue looks to be because Wistia.js is using the Wistia video key as the unique identifier for the array. So two of the same video will not end up with unique entries in the array.

CookPete:
I'm afraid I need more information to look into this. I can render multiple of the same Wistia video and `onPlay` is firing correctly for all of the videos. I can also toggle videos off and on, and onPlay still fires correctly.

brandon-rentpingmedia:
Sorry about that. Our specific use case is we have an overview page with a Wistia video, then you can click into a detailed page with that has a different react component, but the same Wistia video (url). The onPlay event fires fine for the first video, but when you get to the second video, the onPlay does not fire. Similarly, if you start on the detail page (reload browser), the onPlay fires there, but not on the overview. Does that help?

CookPete:
Can you reproduce this in jsFiddle? My attempts to cause the issue to happen aren't working.

jondkinney:
@CookPete I think @brandon-rentpingmedia is correct. If you look here: https://github.com/CookPete/react-player/blob/bb69b1df1972ca50feb97090c8ff35aab5203b3c/src/players/Wistia.js#L113 you can see that the key that is being used for the wistia video is the URL. So if you have two of the same players on the page they are fighting with each other in the virtual dom.

CookPete:
Annoyingly, this is just [how the Wistia player works](https://wistia.com/support/developers/player-api#with-standard-embeds). You render a div with class of `wistia_async_abcde12345` where `abcde12345` is the ID of your video, then call `_wq.push({ id: 'abcde12345', ... })` to embed the player. 

[...]

PR #905, opened 2020-05-31
Add unique wistia player id to have possibility to have multiple players with the same url on one page

MariyaNizovtsova:
Fix the issue #623
Tested on the page with two Wistia players with the same URL and they don't affect each other anymore.

### DIFF
--- a/src/players/Wistia.js
+++ b/src/players/Wistia.js
@@ -1,15 +1,17 @@
 import React, { Component } from 'react'
 
-import { callPlayer, getSDK } from '../utils'
+import { callPlayer, getSDK, randomString } from '../utils'
 import { canPlay, MATCH_URL_WISTIA } from '../patterns'
 
 const SDK_URL = 'https://fast.wistia.com/assets/external/E-v1.js'
 const SDK_GLOBAL = 'Wistia'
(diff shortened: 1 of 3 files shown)


src/players/Wistia.js
```diff
@@ -24,7 +26,7 @@ export default class Wistia extends Component {
     getSDK(SDK_URL, SDK_GLOBAL).then(() => {
       window._wq = window._wq || []
       window._wq.push({
-        id: this.getID(url),
+        id: this.playerID,
```
afzaalahmad, line 29: can we do something like this here:
`id: this.props.id || this.playerID`

src/players/Wistia.js
```diff
@@ -111,7 +113,7 @@ export default class Wistia extends Component {
       height: '100%'
     }
     return (
-      <div key={id} className={className} style={style} />
+      <div id={this.playerID} key={id} className={className} style={style} />
```
afzaalahmad, line 116: same goes here:
`id={this.props.id || this.playerID}`

CookPete:
I've added a `config.wistia.playerId` prop (similar to facebook and twitch) to let you pass in your own `playerId` if you want.

afzaalahmad:
@CookPete appreciated :+1:

PR #914, opened 2020-06-05
add isMuted listener inside onProgress method

devinekadeni:
This will provide the `muted` status of the video player inside `onProgress` props.
The actual case is when user clicking the mute button on the control player, currently we cannot track the `muted` status on it.

Screen Shot 2020-06-06 at 01 54 05

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -114,6 +114,7 @@ export default class ReactPlayer extends React.Component<ReactPlayerProps, any> 
   seekTo(amount: number, type?: 'seconds' | 'fraction'): void;
   getCurrentTime(): number;
   getDuration(): number;
+  getIsMuted(): boolean;
   getInternalPlayer(key?: string): Object;
   showPreview(): void;
 }
--- a/src/Player.js
+++ b/src/Player.js
@@ -95,6 +95,11 @@ export default class Player extends Component {
     return this.player.getDuration()
   }
 
+  getIsMuted () {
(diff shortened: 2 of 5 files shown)


devinekadeni:
@CookPete please review this PR.
Let me know if I miss something or if I could achieve this with another way.
Thanks!

CookPete:
Hey @devinekadeni, I'm not sure `isMuted` really makes sense as part of `onProgress` – it doesn't really related to the progress of the currently playing media. It also seems wrong to add a top level method like `getIsMuted` that is only supported by FilePlayer.

You can achieve what you are after with `getInternalPlayer()` to access the `<video>` instance when playing files:

```jsx
class Player extends React.Component {
  ref = player => {
    this.player = player
  }
  handleProgress = data => {
    console.log({
      ...data,
      muted: this.player.getInternalPlayer().muted
    })
  }
  render () {
    return (
      <div className='player-wrapper'>
        <ReactPlayer
          ref={this.ref}
          url='http://commondatastorage.googleapis.com/gtv-videos-bucket/sample/BigBuckBunny.mp4'
          playing
          controls
          onProgress={this.handleProgress}
        />
      </div>
    )
  }
}
```

https://jsfiddle.net/nxkfue56/

devinekadeni:
ahh I see, thank you for pointing out the mistakes and gave me the solution.
Should have asked first on GH issue instead, my apologies.
I'll close this PR, Thanks!

PR #964, opened 2020-07-07
Make force disabling PIP dependent on prop

foysalit:
This is a great library and thank you for creating and maintaining it @CookPete . I was stoked to see that the library has support PIP but it feels a bit backwards to disable/remove pip when video is unmounted when most of the time, PIP is intended to be fired when the video is not explicitly visible and other content is given priority. However, as the library owner, I fully respect your decision to implement the UX you think is right for the library. So, this PR does not override the behavior you've implemented and adds a new prop to opt-in for that as a feature.


CookPete:
Hey @foysalit, thanks for the PR. 

I like the idea but I've made it a more generic `stopOnUnmount` prop that defaults to `true`, so you can use 

```
pip={true} stopOnUnmount={false}
```

PR #975, opened 2020-07-22
vimeo: listen for bufferring events and handle with matching callbacks

metamorfoso:
Hey there, thanks for the nice player, it's been great for us.

Just wanted to a add a small feature for our use case: being able to use the `onBuffer` and `onBufferEnd` callbacks with the vimeo player.

A very basic change, following the Vimeo spec:
https://developer.vimeo.com/player/sdk/reference#bufferstart
https://developer.vimeo.com/player/sdk/reference#bufferend


CookPete:
Nice one 👍

PR #1005, opened 2020-09-02
Fix for iPad Pro user agent detection

### LINKED ISSUE #1004: Loading issue on iPad Pro only ("request desktop site")
**TL;DR** - The isIOS check is out of date and doesn't handle iPad Pro in desktop mode.

#### Current Behavior

If you navigate to a video shown by react-player on an iPad Pro specifically, it never passes `loading` to fire the `onReady` event. Looking at the network tab you will see a red `206 Partial Content` instead of the normal green `206`.

In our use case where we show a spinner tied to `loading`, it never goes away.

By default the iPad Pro is requesting the **desktop site**.

- iPad Pro “request desktop site” user agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/13.0.5 Safari/605.1.15
- iPad Pro “request mobile site” user agent: Mozilla/5.0 (iPad; CPU OS 13_3_1 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/13.0.5 Mobile/15E148 Safari/604.1

#### Expected Behavior
Video loads but still needs to be clicked to play, as other issues have noted due to changing iOS autoplay settings.

#### Steps to Reproduce

1. Navigate to a page in an iPad Pro
1. Ensure it is in "desktop mode"
1. Video will not load
1. "Request mobile mode"
1. Note the video will now loads successfully

#### Environment
- Behind an auth wall unfortunately.

#### Other Information
- Creating a PR to fix this by implementing the improved isIOS check from here https://stackoverflow.com/a/58065241

taitems:
Closes #1004 - users can now load a video on an iPad Pro in default desktop mode.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
-  "name": "react-player-temp-fork",
-  "version": "2.6.1",
+  "name": "react-player",
+  "version": "2.6.2",
   "description": "A React component for playing a variety of URLs, including file paths, YouTube, Facebook, Twitch, SoundCloud, Streamable, Vimeo, Wistia and DailyMotion",
   "main": "lib/index.js",
   "typings": "index.d.ts",
@@ -24,7 +24,7 @@
   },
   "repository": {
     "type": "git",
-    "url": "https://github.com/taitems/react-player.git"
+    "url": "https://github.com/CookPete/react-player.git"
   },
   "keywords": [
     "react",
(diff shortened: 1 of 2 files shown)


CookPete:
This PR was a bit of a mess (not sure what's going on with the syntax changes?) so I've just fixed it in 136836e. Published in `v2.6.2`.

taitems:
Yep, all good thanks for fixing, I have prettier and/or editorconfig running in VSCode so they have a habit of reformatting files I touch.

That being said there's probably value in setting up an editorconfig or prettier config if you don't already have one?

PR #1044, opened 2020-10-11
Add support for Audius.co embeds

brycedev:
Audius.co is a newer, high quality streaming service. 

This pull request adds support for Audius embeds like so `<ReactPlayer url='https://audius.co/embed/track?id=88754&ownerId=275&flavor=card' />`

Tests have been included. Thank you in advance for your consideration.


cookpete:
I'm normally apprehensive about adding bytes to the package that won't be used, but:

- it looks like a cool platform
- people can use lazy players
- I'm planning a rewrite that should cut the bundle size down
- you clearly put a lot of effort into this

So I would be up for adding this. However, it seems to be broken? https://audius.co/embed/api.js returns a 404 for me, and I can't find any docs about the player API.

brycedev:
Thank you so much for getting back with me! I'm glad you're ultimately interested in getting it integrated.
From my understanding there was scheduled maintenance around the time you tested that embed link. It should be available now. 

Also, there is no embed api documentation yet, but I will make submit a PR for the [documentation](https://audiusproject.github.io/api-docs/#audius-api-docs). It is extremely similar to Soundcloud's Widget API, if that is helpful.

brycedev:
Hello @cookpete ! I wanted to see if there was anyway I could help move this forward and get this merged in. I believe everything should be compatible with your latest release.

Yasamato:
Do I see something wrong here or is this just a plain embed with no controls? At least in the demo it is ignoring every command and does not report any value at all

PR #1084, opened 2020-11-19
fix: add playIcon prop to TS declaration

jariz:
`playIcon` is currently missing from the .d.ts file.


jariz:
@cookpete Any update? 
This PR is fairly straightforward, and I need to use hacky workarounds now to use react-player with the `playIcon` prop w/TS.

cookpete:
Apologies @jariz – this went completely under my radar. Cheers for the PR 👍

Issue #953, opened 2020-07-01
Typescript definitions for all exports


vai0:
#### Current Behavior
Typescript definitions for the individual players (react-player/wistia, react-player/youtube, etc...) are missing.

e.g.,
```typescript
import WistiaPlayer from "react-player/wistia";
...

// JSX element type 'ReactPlayerWistia' does not have any construct or call signatures.ts(2604)
<WistiaPlayer ... /> 
``` 
As a result, I need to either 
1. explicitly add my own type declarations, and paint WistiaPlayer with the broad typedefs of ReactPlayer...
```typescript
// react-player/index.d.ts
declare module "react-player/wistia" {
  import ReactPlayer from "react-player";
  export default ReactPlayer;
}
``` 
which isn't exactly great because now props such as `onBuffered`, `onBufferEnded`, `onEnablePIP`, etc... that WistiaPlayer really doesn't support noop and creates a confusing developer experience,

OR

2. again explicitly add my own type declarations, but completely type out the WistiaPlayer export + props that actually work and don't noop.
```typescript
// react-player/index.d.ts
declare module "react-player/wistia" {
    import * as React from "react";
    import ReactPlayer from "react-player";

    type WistiaConfig = {
        wistia: {
            // https://wistia.com/support/developers/embed-options#options
            options?: {
                ...
            };
            playerId?: string;
        };
    };

    export type ReactPlayerWistiaProps = {
        url?: string | string[];
        playing?: boolean;
        loop?: boolean;
        controls?: boolean;
        volume?: number;
        muted?: boolean;
        playbackRate?: number;
        width?: string | number;
        height?: string | number;
        style?: React.CSSProperties;
        progressInterval?: number;
        light?: boolean | string;
        wrapper?: keyof JSX.IntrinsicElements;
        config?: WistiaConfig;
        className?: string;
        onReady?(player: ReactPlayer): void;
        onStart?(): void;
        onPlay?(): void;
        onPause?(): void;
        onEnded?(): void;
        onError?(
            error: any,
            data?: any,
            hlsInstance?: any,
            hlsGlobal?: any
        ): void;
        onDuration?(duration: number): void;
        onSeek?(seconds: number): void;
        onProgress?(state: { played: number; playedSeconds: number }): void;
    };
    export default class ReactPlayerWistiaPlayer extends React.Component<
        WistiaPlayerProps,
        any
    > {
        static canPlay(url: string): boolean;
        static canEnablePIP(url: string): boolean;
[...]

PR #1098, opened 2020-12-04
Added types for single-type players

JuanM04:
This PR solves #953

What I did:
- Created a folder named `types` with all the typings.
- Added `cp types/*.d.ts .` to `npm run prepublishOnly`, so the types will be detected.
- Added utilities to `types/_utils.d.ts`.
- Added linting with `npm run lint:ts`

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -40,6 +40,7 @@ export interface VimeoConfig {
 export interface WistiaConfig {
   options?: Object;
   playerId?: string;
+  customControls?: any[];
 }
 
 export interface MixcloudConfig {
--- a/src/players/Wistia.js
+++ b/src/players/Wistia.js
@@ -20,7 +20,10 @@ export default class Wistia extends Component {
 
   load (url) {
     const { playing, muted, controls, onReady, config, onError } = this.props
-    getSDK(SDK_URL, SDK_GLOBAL).then(() => {
+    getSDK(SDK_URL, SDK_GLOBAL).then(Wistia => {
+      if (config.customControls) {
+        config.customControls.forEach(control => Wistia.defineControl(control))
+      }
       window._wq = window._wq || []
       window._wq.push({
         id: this.playerID,
--- a/src/props.js
+++ b/src/props.js
@@ -60,7 +60,8 @@ export const propTypes = {
     }),
     wistia: shape({
       options: object,
-      playerId: string
+      playerId: string,
+      customControls: array
     }),
     mixcloud: shape({
       options: object
@@ -164,7 +165,8 @@ export const defaultProps = {
     },
     wistia: {
       options: {},
-      playerId: null
+      playerId: null,
+      customControls: null
     },
     mixcloud: {
       options: {


types/_utils.d.ts
```diff
@@ -0,0 +1,59 @@
...
+  onProgress?: (state: {
+    played: number
+    playedSeconds: number
+    loaded: number
+    loadedSeconds: number
+  }) => void
+  [otherProps: string]: any
+}
+
+export class BaseReactPlayer {
```
afzaalahmad, line 48: In my opinion, instead of repeating all properties of class in every extended player is redundant. You can use generics to remove this redundancy.
You should extend your base class from React.Component and use generics to implement extends base props types.
Following syntax:
```js
export default class BaseReactPlayer<T extends BaseReactPlayerProps> extends React.Component<T, any> {
   // all properties
}
```
Now in every player types use this exported class like this:
```js
export default class DailyMotionPlayer extends BaseReactPlayer<DailyMotionPlayerProps> {
  // any extended properties for dailymotion only.
}
```
In this case, you will not have to extend DailyMotionPlayerProps from BaseReactPlayerProps because generics will extends thos automatically.

types/_utils.d.ts
```diff
@@ -0,0 +1,59 @@
...
+  static canPlay (url: string): boolean;
+  static canEnablePIP (url: string): boolean;
+  static addCustomPlayer (player: ReactPlayer): void;
+  static removeCustomPlayers (): void;
+  seekTo (amount: number, type?: 'seconds' | 'fraction'): void;
+  getCurrentTime (): number;
+  getSecondsLoaded (): number;
+  getDuration (): number;
+  getInternalPlayer (key?: string): Object;
```
afzaalahmad, line 57: Don't use `Object` as a type. Use Record<string, any> instead.

afzaalahmad (changes requested):

types/_utils.d.ts
JuanM04: Thank you. I didn't know how to do it correctly

types/_utils.d.ts
JuanM04: I just copied old types. I'll change it

types/dailymotion.d.ts
```diff
@@ -0,0 +1,11 @@
+import { BaseReactPlayer, BaseReactPlayerProps } from './_utils'
+
+export interface DailyMotionConfig {
+  params?: Record<string, any>
+}
+
+export interface DailyMotionPlayerProps extends BaseReactPlayerProps {
```
afzaalahmad, line 7: To the best of my understanding, you don't have to extend `DailyMotionPlayerProps` from `BaseReactPlayerProps` because in the implementation of BaseReactPlayer, every given Props are being extended automatically.
```js
export class BaseReactPlayer<T extends BaseReactPlayerProps> extends Component<T, any> {
}
```
any interface provided to BaseReactPlayer will extends BaseReactPlayerProps automatically, you don't have to extend them manually.

afzaalahmad (changes requested):

types/_utils.d.ts
```diff
@@ -0,0 +1,59 @@
...
+  onProgress?: (state: {
+    played: number
+    playedSeconds: number
+    loaded: number
+    loadedSeconds: number
+  }) => void
+  [otherProps: string]: any
+}
+
+export class BaseReactPlayer<T extends BaseReactPlayerProps> extends Component<T, any> {
```
afzaalahmad, line 48: This class should be a default export of this class.

afzaalahmad (changes requested):

types/dailymotion.d.ts
JuanM04: That's intentional. If someone wants the props of the (for example) `DailyMotionPlayer`, they will do `import DailyMotionPlayer, { DailyMotionPlayerProps } from 'react-player/dailymotion'`

afzaalahmad: that's a fair point :100:

types/index.d.ts
```diff
@@ -0,0 +1,31 @@
...
+  dailymotion?: DailyMotionConfig
+  vimeo?: VimeoConfig
+  file?: FileConfig
+  wistia?: WistiaConfig
+  mixcloud?: MixcloudConfig
+  vidyard?: VidyardConfig
+  twitch?: TwitchConfig
+}
+
+export type ReactPlayerProps = BaseReactPlayerProps & {
```
afzaalahmad, line 27: this must be interface which should extends `BaseReactPlayerProps` because if you are exporting it then also give the user to extend it. 
Interfaces can be extended but not type.

types/streamable.d.ts
```diff
@@ -0,0 +1,5 @@
+import BaseReactPlayer, { BaseReactPlayerProps } from './base'
+
+export type StreamablePlayerProps = BaseReactPlayerProps
```
afzaalahmad, line 3: Same concern, it must be interface instead of type.

afzaalahmad (changes requested):

afzaalahmad (approved):

afzaalahmad:
@JuanM04 please make sure to use proper commit messages. It would be awesome to use `git interactive rebasing` and merge these all commits into one commit with proper message `added player typings`.

Regards

JuanM04:
@afzaalahmad merging all the commits into one would be awesome. If there is something I can do to help, please tell me

cookpete:
>It would be awesome to use git interactive rebasing and merge these all commits into one com

There's no need as I'm using [squash and merge](https://docs.github.com/en/free-pro-team@latest/github/collaborating-with-issues-and-pull-requests/about-pull-request-merges#squash-and-merge-your-pull-request-commits) in this repo, so the commits will be squashed into one merge commit anyway.

isometriq:
types for custom players would be awesome too

cookpete:
Amazing work 💪    

I managed to break everything when trying to rebase on `master` and force push (accidentally force pushed my `master` rather than my rebased branch of your `master`) so I'll merge this manually.

Issue #508, opened 2018-10-26
Failed to execute 'postMessage' on 'DOMWindow'


Tresky:
I last worked on my project about a month ago. At that time, I was able to get Youtube videos to load and play properly in react-player. After coming back, I've discovered that something seems to have changed and now the video won't load. I end with the following error:
```
Failed to execute 'postMessage' on 'DOMWindow': The target origin provided ('https://www.youtube.com') does not match the recipient window's origin ('http://localhost:8000').
```
I've looked up some StackOverflow questions about it and I've read that it's an error coming from the Youtube iFrame API when you make a request for a Youtube video over http instead of https. I put some logging in the react-player library and saw that the requests are supposedly being made over https, so I don't think that's the issue I'm having here.

Does anyone have any insight into this?

As an aside, I'm using Gatsby as my platform. Not sure if this really matters.

NathanielHill:
I'm getting the same warning although the videos work, did you resolve it?

lloydh:
@Tresky I'm getting this too. Any tips on a fix?

Tresky:
@NathanielHill @lloydh Hey guys. Whenever I was working on this, I discovered that I get the error mentioned above, but the player still functions properly. I never found a solution to stop this error from appearing.

carpben:
I'm getting this error too. 
@CookPete , Are you aware of this issue? 
@Tresky , you closed the issue, even though it's relevant and hasn't been resolved. Do you mind reopening it? 
![image](https://user-images.githubusercontent.com/13496178/51187720-175c1980-18e5-11e9-9283-5b9535ceda32.png)

Tresky:
I've reopened this issue because it seems to be a common problem experienced by many people. Hopefully, we can find a solution here.

Does the issue happen only on page load? Do all postMessage events function properly after the initial page load? It could be that events are trying to be [posted to the hidden iframe before it gets loaded fully](https://stackoverflow.com/questions/22194409/failed-to-execute-postmessage-on-domwindow-the-target-origin-provided-does).

CookPete:
I'm aware of the issue, but afaik it's a problem with with the Youtube iFrame API based on whatever URL you are calling the API from, and it's not very consistent. This issue should remain closed unless there is a specific problem that ReactPlayer is causing.

There are plenty of mentions of the error elsewhere: https://www.google.com/search?q=youtube+iframe+Failed+to+execute+%27postMessage%27+on+%27DOMWindow%27%3A

[...]

PR #1118, opened 2021-01-03
Fix/extract youtube config

iaarnio:
## Description

According to docs youtube config parameters are extracted from youtube key:

```
There is a single config prop to override settings for each type of player:

<ReactPlayer
  url={url}
  config={{
    youtube: {
      playerVars: { showinfo: 1 }
    },
    facebook: {
    ...
```

However, in reality these where extracted from top level.  Therefore user settings were practically ignored.

Note that is is potentially breaking change if some library users have used the youtube configs from top level (against what is instructed documentation).

## Solution

Extract youtube configs under youtube key as documented.

## Notes

I come across this when trying to find solution for an old issue https://github.com/cookpete/react-player/issues/508
To my understanding this can be solved by passing `enablejsapi: 1` as playerVar (with explicit origin too if needed).

However, I was unable to get this working until I figured out that youtube playerVars were extracted differently than documentation explains.

### DIFF
--- a/src/players/YouTube.js
+++ b/src/players/YouTube.js
@@ -29,7 +29,7 @@ export default class YouTube extends Component {
 
   load (url, isReady) {
     const { playing, muted, playsinline, controls, loop, config, onError } = this.props
-    const { playerVars, embedOptions } = config
+    const { youtube: { playerVars, embedOptions } } = config
     const id = this.getID(url)
     if (isReady) {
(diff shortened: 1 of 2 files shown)


cookpete:
As of `v2.0`, you no longer need to specify the `youtube` key in the `config` prop. Both the old style and new style will work, so it was not a breaking change.

https://github.com/cookpete/react-player/blob/master/MIGRATING.md#the-config-prop

PR #1122, opened 2021-01-04
Add missing FLV properties to TypeScript definitions

rmsy:
There are a couple of FLV properties﻿ missing from `index.d.ts` -- this adds them.


cookpete:
Thanks 👍

PR #1133, opened 2021-01-14
FEAT: customize fallback through props for lazy loading

afzaalahmad:
- Added props for providing customized fallback in case of lazy loading.


cookpete:
Nice work!

PR #1145, opened 2021-01-20
fix: nullify srcObject instead of calling removeAttribute

komachi:
Sorry, I was wrong in #1121, as srcObject is not a content attrubute, but an IDL attribute, it cannot be cleared with removeAttribute. This pr fixes this and properly nullify srcObject [as per spec](https://html.spec.whatwg.org/multipage/media.html#dom-media-srcobject).


src/players/FilePlayer.js
```diff
@@ -38,7 +38,7 @@ export default class FilePlayer extends Component {
       this.props.url !== prevProps.url &&
       !isMediaStream(this.props.url)
     ) {
-      this.player.removeAttribute('srcObject')
+      this.player.srcObject = null;
```
cookpete, line 41:
```suggestion
      this.player.srcObject = null
```

cookpete (changes requested):

PR #1148, opened 2021-01-21
Remove tabIndex from preview component

edunad:
Hi!

Currently developing a project that uses react-player and we are using a custom focus system. Having this `tabIndex` on the preview is sadly breaking it, and there is no easy way to remove it.

I think the `tabIndex` should be handled by the user implementing the player (Aka : wrap it around a div with tabindex)

### DIFF
--- a/src/Preview.js
+++ b/src/Preview.js
@@ -95,7 +95,6 @@ export default class Preview extends Component {
         style={styles.preview}
         className='react-player__preview'
         onClick={onClick}
-        tabIndex={0}
         onKeyPress={this.handleKeyPress}
       >
         {playIcon || defaultPlayIcon}


cookpete:
>I think the tabIndex should be handled by the user implementing the player (Aka : wrap it around a div with tabindex)

Wouldn't that prevent the user from being able to "click" the preview using the keyboard? You would need the clickable thing in focus.

I'm happy to add a `previewTabIndex` prop that defaults to `0`? Or maybe just `previewProps` that lets you override anything on that div.

edunad:
> I'm happy to add a previewTabIndex prop that defaults to 0? Or maybe just previewProps that lets you override anything on that div.

That's a better solution I think

PR #1153, opened 2021-01-23
fix: ensure lazy typedefs mirror defaults

### LINKED ISSUE #1146: TypeScript declarations for react-player/lazy broken in v2.8.1
#### Current Behavior
Previously successfully using react-player v2.7.2 in TypeScript project importing react-player/lazy.

Upgrading to react-player v2.8.1 results in TypeScript errors as the TypeScript declarations for react-player/lazy have been broken.

#### Expected Behavior
Working TypeScript declarations for react-player/lazy.

#### Steps to Reproduce
Just try importing "react-player/lazy" in TypeScript project using react-player v2.8.1.

#### Environment
- URL attempting to play: n/a
- Browser: Chrome
- Operating system: Windows 10
- jsFiddle example: n/a

#### Other Information
Related to old closed issue https://github.com/cookpete/react-player/issues/918.

The react-player/lazy/index.d.ts file seems to be a copy of the react-player/index.d.ts file but this simple approach no longer works due to broken relative path imports.

jahed:
Fixes #1146

## Notes

~~To avoid the cyclic dependency (which would cause an invalid lazy -> base -> lib dependency), since base.d.ts is used internal only, it's safe to move the default types there (to avoid base -> lib) and re-export them in `index.d.ts`.~~

I also moved `index.d.ts` to `lib/index.d.ts` as that's where `lib/index.js` is. So `main` and `typings` match to avoid confusion. This also solves the relative path issue when copying it for `lazy/index.d.ts`.


JuanM04:
> I also moved index.d.ts to lib/index.d.ts as that's where lib/index.js is. So main and typings match to avoid confusion. This also solves the relative path issue when copying it for lazy/index.d.ts.

That's pretty neat

> To avoid the cyclic dependency (which would cause an invalid lazy -> base -> lib dependency), since base.d.ts is used internal only, it's safe to move the default types there (to avoid base -> lib) and re-export them in index.d.ts.

You just changed where the cyclic dependency happens: now, for example,`soundcloud.d.ts` imports `base.d.ts` and vice-versa.
Maybe, you should create a `types/lazy/index.d.ts` like this:

```ts
export * from '../index'
export { default } from '../index'
```

jahed:
>You just changed where the cyclic dependency happens

Ah, of course. 🤦🏽 I'll avoid resolving the cyclic dependency. Thought it was causing errors so went for a quick "fix" but it was actually `npm link` anyway. (https://github.com/Microsoft/TypeScript/issues/6496#issuecomment-384786222)

I've pushed an update.

types/lazy/index.d.ts
```diff
@@ -0,0 +1,2 @@
+export * from '../lib'
+export { default } from '../lib'
```
cookpete, line 2: Newline at the end of this file would match the rest of the codebase 👍

cookpete (changes requested):

types/lazy/index.d.ts
jahed: Yeah, I saw that. Pushing again.

package.json
```diff
@@ -21,7 +21,7 @@
     "build:standalone": "cross-env NODE_ENV=production webpack --config webpack/standalone.babel.js",
     "preversion": "npm run lint && npm run test",
     "version": "auto-changelog -p && npm run build:dist && npm run build:standalone && git add CHANGELOG.md dist",
-    "prepublishOnly": "npm run build:lib && npm run build:lazy && npm run build:dist && node scripts/pre-publish.js && cp types/*.d.ts . && cp types/index.d.ts lazy",
+    "prepublishOnly": "npm run build:lib && npm run build:lazy && npm run build:dist && node scripts/pre-publish.js && cp types/*.d.ts . && cp types/lib/index.d.ts lib && cp types/lazy/index.d.ts lazy",
```
cookpete, line 24: Would this work?

```suggestion
    "prepublishOnly": "npm run build:lib && npm run build:lazy && npm run build:dist && node scripts/pre-publish.js && cp -r types/* .",
```

JuanM04: I was thinking about that. I just tested it in my Linux machine and it works

jahed: I was worried it might overwrite directories. But it does work. 👍🏽

cookpete (approved):

cookpete:
Great work @jahed. Thanks for the help @JuanM04 👍

PR #1169, opened 2021-02-08
Adds previewTabIndex property to light mode

edunad:
Tested using local build


cookpete:
Nice work! Note that this fixes https://github.com/cookpete/react-player/pull/1148

Issue #1174, opened 2021-02-16
Vimeo: race condition between `ready` state and call `seekTo`


newsiberian:
Be sure to [search for your issue](https://github.com/CookPete/react-player/issues) before opening a new one.

#### Current Behavior

When users call `seekTo` for some services like **Vimeo** they could receive an error `ReactPlayer: could not seek using fraction – duration not yet available` because `seekTo` is depends on video `duration`. Receiving `duration` in case of **Vimeo** is async and it could happen a little bit later after `loaded` will be received.

#### Expected Behavior

`ready` state must be emitted after all required parameters will be received (at least `duration`)

#### Steps to Reproduce

You need to store `played` state somewhere locally and use it to call `seekTo` right after player will emit `ready` and also you must use pass `playing: true` to the player

1. Set progress to the middle of the video
1. Refresh page

cookpete:
I’m not sure what ReactPlayer can/should do about this. If you are seeking using fractions I would suggest disabling seeking in your app until `onDuration` has been called. Waiting for duration before calling `onReady` would couple the two events to tightly and just isn't necessary. Use `onDuration` rather than `onReady` if you require duration for your UI to be "ready".

newsiberian:
Hi, @cookpete. I dont think you are right here. Library authors should strive to provide one API for all such things like different media services (in our case). So, I, like library user, don't wont to write different logic for each service (youtube, vimeo, etc) on my side

cookpete:
>Library authors should strive to provide one API for all such things.

I agree with this, so I’m not sure I understand your point. `onReady` is called when the player is ready to be played. `onDuration` is called when the duration of the video is available. The behaviour of both are consistent across all players.

`onReady` is not "when seeking is ready" – it is "when the video can be played". It doesn’t (and shouldn’t) [care](https://en.wikipedia.org/wiki/Separation_of_concerns) about whether duration is ready or not. That’s what `onDuration` is for. If an action requires duration, you need to block/disable it until `onDuration` is called.

PR #1175, opened 2021-02-16
Fix 1174: race condition between `ready` and `seekTo` for Vimeo

newsiberian:
Wait till `duration` will be received before emit `ready`


cookpete:
See https://github.com/cookpete/react-player/issues/1174#issuecomment-780919990

`onReady` should be called when the player is ready to be played, not when the player has a duration available (which could be much later). If you want to wait until duration is available to do something, use `onDuration`.

PR #1186, opened 2021-03-04
don't stack event listeners

thomas-short:


cookpete:
Published in `2.10.1`

PR #1213, opened 2021-04-16
update video extenstions to accept media fragments for time

dednal:
Add support for supplying urls that contain media fragments for time
https://www.w3.org/TR/media-frags/#naming-time

example:
https://test-videos.co.uk/vids/bigbuckbunny/mp4/h264/360/Big_Buck_Bunny_360_10s_1MB.mp4#t=5
the video will start playing at 5 seconds

Currently if you directly import the file player if won't load these types of urls
you can get around this by using the lazy import and it will fallback to the file player.

This change will allow for the direct import of react-player/file


cookpete:
Published in `2.10.0`

PR #1223, opened 2021-04-27
Handle setPlaybackRate error

bravecow:
setPlaybackRate can throw error:

```
DOMException: Failed to set the 'playbackRate' property on 'HTMLMediaElement': The provided playback rate (20) is not in the supported playback range.
```

For now this exception is unhandled.

It needs to be inside try / catch:

https://googlechrome.github.io/samples/media/playback-rate-exception


cookpete:
Published in `2.10.0`

PR #1229, opened 2021-05-10
add title attribute to vimeo iframe

danhemerlein:
This adds `title` as a supported prop for `<ReactPlayer />`.  When `<ReactPlayer />` is a `<Vimeo />` component that title will appear as a title attribute on the `iframe` within the container.

Having a title attribute on an `iframe` is critical for screen reader users to tell them what is contained within that element.

more info on this topic from W3 can be found [here](https://www.w3.org/TR/WCAG20-TECHS/H64.html#:~:text=The%20objective%20of%20this%20technique,enter%20and%20explore%20in%20detail.&text=The%20iframe%20element%20remains%20part%20of%20the%20HTML5%20specification)

### DIFF
--- a/src/players/Vimeo.js
+++ b/src/players/Vimeo.js
@@ -36,6 +36,7 @@ export default class Vimeo extends Component {
         const iframe = this.container.querySelector('iframe')
         iframe.style.width = '100%'
         iframe.style.height = '100%'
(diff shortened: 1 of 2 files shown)


sangmshin (approved):

cookpete:
Published in `2.10.0`

PR #1230, opened 2021-05-12
fix: don't defer MediaStream

komachi:
There is no reason to defer MediaStream loading, also this doesn't work because `onReady` is never called for MediaStream as srcObject is inserted later.


cookpete:
Published in `2.10.0`

PR #1234, opened 2021-05-19
Add type for wrapper in base.d.ts

Sky161:


cookpete:
Published in `2.10.0`

alessandroliveri:
@cookpete does this mean that wrapper is not supposed to support `string` anymore?

rhwinter:
Hi there @cookpete,
isn't the `wrapper` prop missing some stuff:
https://github.com/cookpete/react-player/blob/master/src/ReactPlayer.js#L175
?

As an example when I do something like
```
import ReactPlayer from "react-player";
...
<ReactPlayer
    ...
    wrapper={({ children, style }) => <div style={style}>{children}</div>}
/>
```

I get a Typescript error:
```
Property 'style' does not exist on type '{ children: ReactNode; }'.ts(2339)
```

In [the code](https://github.com/cookpete/react-player/blob/795b19614977fbe2b89f6fd14503d1bfb121a722/src/ReactPlayer.js#L175) I see that `<Wrapper>` is:
```
<Wrapper ref={wrapperRef} style={{ ...style, width, height }} {...attributes}>
```

I don't know about the other things (`ref` and `{...attributes}`), but `style` should be `CSSProperties`, right?

So in [`types/base.d.ts`](https://github.com/cookpete/react-player/blob/795b19614977fbe2b89f6fd14503d1bfb121a722/types/base.d.ts#L42) there could be something like:
```diff
@@ -39,7 +39,7 @@ export interface BaseReactPlayerProps {
  stopOnUnmount?: boolean
  light?: boolean | string | ReactElement
  fallback?: ReactElement;
- wrapper?: ComponentType<{children: ReactNode>;
+ wrapper?: ComponentType<{children: ReactNode, style: CSSProperties}>;
  onReady?: (player: ReactPlayer) => void
  onStart?: () => void
  onPlay?: () => void
```

Should I make a pull request for this?

Thanks a lot for the amazing component!

PR #1256, opened 2021-06-10
Ability to pass null to previewTabIndex

edunad:
I forgot to update the typings when fixing the previewTabIndex on #1169 (my bad)


cookpete:
Published in `2.10.0`

PR #1269, opened 2021-06-19
Add media attribute to source tag in file player

marcus-aemilius:
When using react-player's file player in a typescript project, I noticed that the media attribute is missing in the type definitions. (When working around the types, the media attribute shows up in the `<video>` element due to the spreading operation in the [code](https://github.com/cookpete/react-player/blob/master/src/players/FilePlayer.js#L317).)

Although, unfortunately, the media attribute [seems to be ignored by some browsers](https://stackoverflow.com/questions/25907930/chrome-not-respecting-video-source-inline-media-queries), it's still [part of HTML 5](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/source). I am therefore suggesting to add it to the type definitions and fix the functionality by the test I extended.

### DIFF
--- a/test/players/FilePlayer.js
+++ b/test/players/FilePlayer.js
@@ -343,6 +343,7 @@ test('render - string array', t => {
 
 test('render - object array', t => {
   const url = [
+    { src: 'file.mp4', type: 'video/mp4', media: '(max-width:800px)' },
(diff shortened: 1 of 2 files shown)


cookpete:
Published in `2.10.0`

Issue #161, opened 2017-02-23
Add support for Spotify & Google Music


samuelbriole:
Hi, 

I'm interested in streaming Spotify and Google Music tracks with this library.
Is this in any of your plans ?
Do you have any advise if I want to develop the feature ?

Cheers, Samuel

CookPete:
The closest thing we have for Spotify is the [Spotify Play Button](https://developer.spotify.com/technologies/widgets/spotify-play-button/), which is just a very basic iframe player. All we can do is render an iframe and no other `ReactPlayer` props/interactions would work. They say an API is on the way:

>The Play button doesn’t have any observable events, nor does it listen for any. Playback can only be triggered by the user clicking the button.
>
>We’re certainly looking at creating a JS SDK; it’s a bigger project and right now we’re focusing on quickly releasing APIs to open up the general Spotify ecosystem (catalogue, playlists, etc).
>
>None of these requests are falling on deaf ears, and we’re working hard to bring out as much functionality as we can.

Is it worth adding "support" for Spotify URLs even if it involves just rendering an iframe with no interactions/events?

Also, it doesn't looks like Google Music has an API we can use. All I can find is unofficial proxies/hacks.

samuelbriole:
@CookPete In the meantime I also looked at which APIs we could use, and I got the same results as you.

I hope Spotify is indeed working on a JS SDK :)

samuelbriole:
@CookPete 

It looks like we have something promising for Spotify https://beta.developer.spotify.com/documentation/web-playback-sdk/

CookPete:
@samuelbriole Good spot! I'll look into adding support for this.

rasjonell:
hey @CookPete, we've been using react-player for some time and we're really happy with it. I came across this issue and saw that there is a branch for adding Spotify support and was wondering what state it is in? will it be merged into master any time soon?

vincentwinkel:
Some update about the spotify integration ?
I'm using react-player for youtube, soundcloud etc, and I'm trying to integrate spotify aside. Not so easy because of the auth, but as fallback their sdk provides a `preview_url` (the 30s free snapshot). Unfortunately, even those urls cannot be played using react-player.
Eg. https://p.scdn.co/mp3-preview/0ed0e3241d7dd381271156710b7153308ddfbb89

PR #1316, opened 2021-09-15
Add support to embed Spotify player

StevenHoppinger24K:
I have a request for adding spotify support, I noticed there was allready support for soundcloud and mixcloud, but no spotify. 

I have added a new player class and construct the url to embed spotify right away. 
https://developer.spotify.com/documentation/widgets/guides/adding-a-spotify-podcast-embed/

Cause it is only a matter of adding the word /embed/ in front of /track/<ID> or /playlist/<ID> or /episode/<ID> for podcasts. 

I did not implement the getSdk method I found in there cause I could not find a spotify SDK that can be used without siging in.


StevenHoppinger24K:
@cookpete, What do think of this? Spotify is used a lot for podcasts. So I think this could be a valueable standard feature.

cookpete:
What is the benefit if the player is not interactive and does not emit events for us to listen to? You would be including the entire `react-player` bundle just to render something like

```html
<iframe src="https://open.spotify.com/embed/album/1DFixLWuPkv3KT3TnV35m3" width="300" height="380" frameborder="0" allowtransparency="true" allow="encrypted-media"></iframe>
``` 

which you could just paste directly into your app.

cookpete:
See also:
- https://github.com/cookpete/react-player/pull/839#issuecomment-604651779
- https://github.com/cookpete/react-player/issues/161#issuecomment-296483444

PR #1333, opened 2021-10-17
Add oEmbedUrl as changeable prop

luwes:
This change makes it possible to use another service for retrieving the preview image.

I tested the code by adding this property to `ReactPlayer` in App.js:

```js
oEmbedUrl='https://api.playerx.io/oembed'
```

Noembed doesn't seem to be maintained anymore and there is no purge cache support and the default cache is set to expire after 100 days.

I started a new oEmbed service with purge support over at https://github.com/playerxo/oembed
It's hosted on Cloudflare workers which offers a generous free tier and it's written in Javascript.
Feel free to use it or host your own with the pubic code.

### DIFF
--- a/src/Preview.js
+++ b/src/Preview.js
@@ -26,7 +26,7 @@ export default class Preview extends Component {
     this.mounted = false
   }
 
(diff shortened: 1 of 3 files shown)


cookpete:
Published in `2.10.0`

PR #1354, opened 2021-11-14
Add onSeek support to Twitch player

romain-daversa:
Hello. 
First off thanks for this amazing player. 
Juste a little commit to add support of "onSeek" event on the Twitch player since it's now supported by the embed API ([see here](https://dev.twitch.tv/docs/embed/video-and-clips#javascript-events))


cookpete:
Published in `2.10.0`

PR #1360, opened 2021-11-26
(fix): clear video src to prevent old video from continue loading

### LINKED ISSUE #1359: Changing url, previous video continue loading
#### Current Behavior
When using `FilePlayer`,  changing `url`, the old video still continue to load (can be observe in the network request), new `url` in pending state until the old video finish loading. 

#### Expected Behavior
Previous video stop loading, new url load immediately

#### Environment
- URL attempting to play:  N/A
- Browser: Chrome
- Operating system: Mac 
- jsFiddle example: N/A

ptgamr:
Fix #1359

Inspired by prior-art from videogular:

https://github.com/videogular/videogular/blob/master/app/scripts/com/2fdevs/videogular/controllers/vg-controller.js#L698

### DIFF
--- a/src/players/FilePlayer.js
+++ b/src/players/FilePlayer.js
@@ -43,6 +43,7 @@ export default class FilePlayer extends Component {
   }
 
   componentWillUnmount () {
+    this.player.src = ''
     this.removeListeners(this.player)
     if (this.hls) {
       this.hls.destroy()


novyQ:
Thanks for fixing it. we had the same issue that the player keeps loading old url.

ptgamr:
@novyQ cheers, now just waiting for someone to merge it, hopefully

novyQ:
> @novyQ cheers, now just waiting for someone to merge it, hopefully

noticed the CI error. is that why the PR is not being merged?

src/players/FilePlayer.js
```diff
@@ -38,11 +38,13 @@ export default class FilePlayer extends Component {
       this.props.url !== prevProps.url &&
       !isMediaStream(this.props.url)
     ) {
+      this.player.src = ''
```
cookpete, line 41: I don’t think this is needed? This runs when the `url` has changed and the new one is not a MediaStream, we don’t necessarily want to clear the `src` here, especially if the new `url` is also a file path?

ptgamr: @cookpete I have tested it without this line and it doesn't really fix the problem (old video still load when I change the URL)

I don't see any downside of clearing `src` here, as it'll be set again in the next render.

cookpete: >I don't see any downside of clearing src here, as it'll be set again in the next render.

Will it? This is in `componentDidUpdate` which fires after everything has finished rendering. When does the "next render" happen?

ptgamr: @cookpete Oh you're right. Sorry, I was blind and somehow thought it is `componentWillUpdate`... Now, I don't even know why it's still working when I have this. 

I undo this change, and looks like reseting the src inside `componentWillUnmount` is enough in my case (as I actually unmount the videoPlayer)

Thanks for looking into this!

cookpete:

PR #1372, opened 2021-12-21
fix: Add exception handling when seekTo is called with amount value 0

imjihun:


cookpete:
Published in `2.10.0`

PR #1393, opened 2022-02-04
Make pre-publish.js ES5 compatible

unformatt:
`pre-publish.js` injects code that circumvents webpack's `target` option. Meaning, if you specify in webpack `target = ['web', 'es5']` and are expecting an es5-compatible bundle, the bundle will be broken because of the `const` injected here; `const` is not es5.


cookpete:
Published in `2.10.0`

Issue #1320, opened 2021-09-23
Passing component to light param instead of image url


andrewscofield:
I'm working on a next.js project, where images are automatically sized and optimized on the backend. Being able to pass a Component to the light parameter would be amazing to be able to use those images to help optimize between mobile and desktop.

idiglove:
@andrewscofield What attributes would those components have? Would they be img tags or something else? Can you elaborate more? Thanks

cookpete:
Fixed by https://github.com/cookpete/react-player/pull/1405 I think?

andrewscofield:
Yes, this looks awesome! Thanks @idiglove for working on this and @cookpete for getting it merged and letting me know about it.

PR #1405, opened 2022-02-24
ability to pass a component in light prop

idiglove:
the light prop could be like this:
`light={sun}`

The styling of the image component should come from where it was called

The play icon will be positioned absolute if `light` is a component 

image

refers to #1320

### DIFF
--- a/src/Preview.js
+++ b/src/Preview.js
@@ -7,7 +7,8 @@ const cache = {}
 export default class Preview extends Component {
   mounted = false
   state = {
-    image: null
+    image: null,
+    previewImage: { isComponent: false, component: false }
   }
 
   componentDidMount () {
@@ -30,6 +31,10 @@ export default class Preview extends Component {
     if (typeof light === 'string') {
       this.setState({ image: light })
       return
+    }
+    if (typeof light !== "string" && typeof light !== "boolean") {
+      this.setState({ previewImage: { isComponent: true, component: light } });
+      return;
     }
     if (cache[url]) {
       this.setState({ image: cache[url] })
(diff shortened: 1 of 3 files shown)


src/Preview.js
```diff
@@ -31,6 +32,10 @@ export default class Preview extends Component {
       this.setState({ image: light })
       return
     }
+    if (typeof light !== "string" && typeof light !== "boolean") {
+      this.setState({ previewImage: { isComponent: true, component: light } });
+      return;
+    }
```
n3llym, line 38: I'm not sure it's a good idea to assume than if it isn't a string or boolean then it is definitely a component.  Maybe you could try 
```
 else if (React.isValidElement(light) {
        this.setState({ previewImage: { isComponent: true, component: light } });
	return;
}
```

src/Preview.js
```diff
@@ -99,6 +106,7 @@ export default class Preview extends Component {
         onKeyPress={this.handleKeyPress}
       >
         {playIcon || defaultPlayIcon}
+        {previewImage.isComponent && previewImage.component}
       </div>
```
n3llym, line 110: If you switch the order of these you don't need to add the z-index to the shadow style.
i.e 
```
{previewImage.isComponent && previewImage.component}
{playIcon || defaultPlayIcon}
```

n3llym, line 110: If you switch the order of these you don't need to add the z-index to the shadow style.
i.e 
```
{previewImage.isComponent && previewImage.component}
{playIcon || defaultPlayIcon}
```

n3llym (commented):
Thanks @idiglove, and good job. It would be great to get this added!

Just a few things I noticed I've added some inline comments as well as the below.

1. You need to add the new type in base.d.ts i.e `light?: boolean | string | ReactElement`
2. You need to update the light prop type in props.js i.e` light: oneOfType([bool, string, object])`
3. Also don’t forget to run lint and run the tests!

idiglove:
Thanks @n3llym I've done the changes, linted and tested. Appreciate your review!

cookpete:
I tidied things up a bit to remove the need for the extra stuff in `state`. Nice work!

cookpete:

PR #1415, opened 2022-03-10
Added .m4b extension to AUDIO_EXTENSIONS

s00nice:
the .m4b extension is identical to a .m4a extension with extra audiobook related metadata. m4b plays without issues in ReactPlayer


cookpete:
Published in `2.10.0`

PR #1425, opened 2022-03-27
Generate a file `dist/ReactPlayer.standalone-module.js` ES6 module

pikilon:
A simple script to import `renderReactPlayer` as an ES6 module instead of polluting the global scope.


pikilon:
I'm 45, it is my first commit in a repository I use.
Very happy about it

cookpete:
Nice job @pikilon! It’s never too late 😄

cookpete:

PR #1432, opened 2022-04-06
Allow parameters after entry_id on kaltura

dottodot:
Currently the url provided must end with an entry_id parameter, which may not always match, for example the url generated in the Drupal media has additional parameters after the entry id.


cookpete:
Published in `2.10.1`

PR #1450, opened 2022-05-05
fixes react-18 youtube error #1439

### LINKED ISSUE #1439: Warnings with React 18
#### Current Behavior
Getting infinite warnings could not call getDuration, getCurrentTime, getVideoLoadedFraction on React 18.
This is only happening when React 18 with createRoot is used.
The app works fine, the video works fine only these infinite warnings are the issue.

image

#### Expected Behavior
Shouldn't have any warnings or any function calls.

#### Steps to Reproduce

1. Codesandbox Link : https://codesandbox.io/s/snowy-field-u9fugp?file=/src/App.js

#### Environment
- React 18 with createRoot

manishsharma004:
fixes #1439

### DIFF
--- a/src/Player.js
+++ b/src/Player.js
@@ -53,7 +53,10 @@ export default class Player extends Component {
       this.isLoading = true
       this.startOnPlay = true
       this.onDurationCalled = false
-      this.player.load(url, this.isReady)
+      if (!this.loaded?.[url]) {
+        this.player.load(url, this.isReady)
+        this.loaded = { [url]: true }
+      }
     }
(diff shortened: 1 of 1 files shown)


lehnerpat:
just tested this by patching my react-player (v2.10.0) contents in `node_modules`, and can confirm that it resolves the problems from #1439.

Would appreciate it if this could be merged and released soon :)

cookpete:
I went for a slightly more elegant fix :) Thanks for the PR though! Fixed in `2.10.1`

PR #1465, opened 2022-06-03
Fix onProgress not firing in Strict Mode

### LINKED ISSUE #1453: [v2.10.1] `onProgress` callback is not firing when the video is playing
Be sure to [search for your issue](https://github.com/CookPete/react-player/issues) before opening a new one.

#### Current Behavior
`onProgress` callback is not called when the video is playing.

#### Expected Behavior
`onProgress` callback should be called periodically with the playback information when the video is playing.

#### Steps to Reproduce
1. Use version `v2.10.1` of react-player (issue does not exist in `v2.10.0`
1. Add a simple player as follows that plays a webm file

    ```jsx
	<ReactPlayer
      ref={playerRef}
      width={1024}
      height={768}
      muted
      playing={true}
      url="https://example.com/video.webm"
      onDuration={onDuration}
      onProgress={onProgress}
    />
	```
1. The `onProgress` callback is never called

#### Environment
- URL attempting to play: Any url containing a video file (I used a webm file)
- Browser: Chrome
- Operating system: MacOS

Taar:
- Update the Player handlePlayerMount method so that the progress
  method is always called.

Should fix #1453 

let me know if that test makes sense or not.

### DIFF
--- a/src/Player.js
+++ b/src/Player.js
@@ -90,7 +90,10 @@ export default class Player extends Component {
   }
 
   handlePlayerMount = player => {
-    if (this.player) return // Prevent loading twice in strict mode
+    if (this.player) { // Prevent loading twice in strict mode
+      this.progress()
+      return
+    }
     this.player = player
     this.player.load(this.props.url)
     this.progress()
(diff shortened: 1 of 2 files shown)


cookpete:
Makes sense! Thanks.

cookpete:

PR #1467, opened 2022-06-16
feat: allow auto play on mixcloud player

JanStevens:
Allow auto playing on mixcloud player, bringing it in line with soundcloud


cookpete:

PR #1475, opened 2022-07-13
Feature/add classname props

mixelburg:
Added proper className prop support


cookpete:
Any additional props passed in to ReactPlayer are added as `{...attributes}` here. So I think `className` is already supported and this PR is redundant? Correct me if I’m wrong.

https://github.com/cookpete/react-player/blob/9775bb755af195e6f7e8bacfed436fcefb135354/src/ReactPlayer.js#L173

mixelburg:
It is but the problem is that it isn't in the typescript docs

mixelburg:
But yea, you are right, I think we should just add it to the typescript definition files

cookpete:
Does the `[otherProps: string]: any` not cover it sufficiently? I guess it would be nice to be specific, but then you could say the same thing about any other possible "other" prop.

PR #1516, opened 2022-10-08
Extending valid DailyMotion URL regex  Fix #1430

### LINKED ISSUE #1430: Issue on Dailymotion link regex
#### Current Behavior
This valid link is not recognized as a Dailymotion link and the React Player fallback on the video element:

http://www.dailymotion.com/video/xdni1e_l-histoire-etrange-1_creation#.UXRVsbXwn90

#### Expected Behavior
Ignore dot or # character in the Dailymotion regex, to display it correctly.

egon-isnoise:
The main issue was that DailyMotion videos would not show in the player,
the technical reason was a regex in pattern.js that would not include the '#.' chars from the URL inserted in the component.

We extended the Regex code to include the '#.' chars.


Maxou44:
Looks good :)

cookpete:
Published in `v2.11.1`

Issue #1466, opened 2022-06-12
Problem launching the second video on an MP4 format


thomaslasky:
#### Current Behavior
When i send an array of URL with MP4 path the player read the first MP4 path, but at the end it don't switch to the second URL.
It work with Youtube URL, but not with MP4 url.

#### Expected Behavior
After read the first MP4 URL Video, it switch to the second one auto

#### Steps to Reproduce
1. Create the player with an array of 2 URL with MP4 files

#### Environment
- URL attempting to play: url={["http://commondatastorage.googleapis.com/gtv-videos-bucket/sample/BigBuckBunny.mp4", "http://commondatastorage.googleapis.com/gtv-videos-bucket/sample/ForBiggerJoyrides.mp4"]} it just sample url just for test my app
- Browser: Firefox 102.0b6
- Operating system: Windows 11 (last version)
- jsFiddle example: https://jsfiddle.net/0wz8nred/4/

#### Other Information
It work perfectly with youtube URL

hernandiazz9:
Hi there, I'm working on this issue, and I find one way to do it. 
I will do a Pull Request and hopefully will be fixed soon. 

Cheers

RobertoLlopis:
PR Code: #1518

PR #1518, opened 2022-10-09
Fix problem launching the second video on mp4 format. Issue: #1466

RobertoLlopis:
Hello again. 

We manage to set logic in FilePlayer in order to reproduce multiple videos and audios in Array passed to url prop.
We tested it with mp4 files.

We found kind of unclear what maps of source tags in the end of the file is for. It was returning source with src as undefined when multiple before.

Feel free to correct or suggest any further consideration.

It passed lint and tests

### DIFF
--- a/src/players/FilePlayer.js
+++ b/src/players/FilePlayer.js
@@ -20,6 +20,11 @@ export default class FilePlayer extends Component {
   static displayName = 'FilePlayer'
   static canPlay = canPlay.file
 
+  state = {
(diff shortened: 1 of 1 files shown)


cookpete:
Custom playlist functionality is outside of the scope of this library. URL arrays in `FilePlayer.js` are used for [multiple sources](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/source#video_example) (eg different formats) of the same video.

Issue #1443, opened 2022-04-22
import ReactPlayer from 'react-player/lazy' fails with vite


DouglasDev:
#### Current Behavior
The app crashes with the following error:

> Uncaught Error: Element type is invalid. Received a promise that resolves to: [object Object]. Lazy element type must resolve to a class or function.
    at mountLazyComponent (react-dom_client.js?v=53134127:14671:17)
    at beginWork (react-dom_client.js?v=53134127:15653:22)
    at beginWork$1 (react-dom_client.js?v=53134127:18882:22)
    at performUnitOfWork (react-dom_client.js?v=53134127:18358:20)
    at workLoopSync (react-dom_client.js?v=53134127:18298:13)
    at renderRootSync (react-dom_client.js?v=53134127:18277:15)
    at recoverFromConcurrentError (react-dom_client.js?v=53134127:17900:28)
    at performConcurrentWorkOnRoot (react-dom_client.js?v=53134127:17848:30)
    at workLoop (react-dom_client.js?v=53134127:195:42)
    at flushWork (react-dom_client.js?v=53134127:174:22)
mountLazyComponent @ react-dom_client.js?v=53134127:14671
beginWork @ react-dom_client.js?v=53134127:15653
beginWork$1 @ react-dom_client.js?v=53134127:18882
performUnitOfWork @ react-dom_client.js?v=53134127:18358
workLoopSync @ react-dom_client.js?v=53134127:18298
renderRootSync @ react-dom_client.js?v=53134127:18277
recoverFromConcurrentError @ react-dom_client.js?v=53134127:17900
performConcurrentWorkOnRoot @ react-dom_client.js?v=53134127:17848
workLoop @ react-dom_client.js?v=53134127:195
flushWork @ react-dom_client.js?v=53134127:174
performWorkUntilDeadline @ react-dom_client.js?v=53134127:382
Show 11 more frames

#### Expected Behavior
I expect the video to load

#### Steps to Reproduce

1. Download my minimal example: https://github.com/DouglasDev/vite-react-player-bug
2. run `npm i` 
3. run `npm run dev`
4. The app will crash

#### Environment
- URL attempting to play: any
- Browser: chrome v 100
- Operating system: mac

cookpete:
It seems like my implementation of lazy loading players is [prone to errors](https://github.com/cookpete/react-player/issues?q=is%3Aissue+lazy). I am open to better suggestions.

[...]

PR #1537, opened 2022-11-10
add unbundled es6 build with jsx file extensions

DouglasDev:
Fixes: https://github.com/cookpete/react-player/issues/1443

This command copies the src files to a directory called 'es6', excluding the demo directory, and renames 'js' files to 'jsx' files if they use jsx. This can then be imported into a project that uses vite this way:

```js
import ReactPlayer from 'react-player/es6'; 
```

This should allow lazy imports as well. 

Perhaps the readme should be updated? Not sure what exactly to write.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,5 +7,6 @@ package-lock.json
 /lazy
 /demo
 /coverage
+/es6
 .idea/
 .vscode/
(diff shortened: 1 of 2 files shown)


solidiquis:
Excited to see this merged. Just came across this problem while migrating from webpack to vite.

wallw-teal:
This is good, but would also like to see a `module` or `browser` field in `package.json` which points to `es6/<entry>`, and then you should be able to use:

```javascript
import ReactPlayer from 'react-player'
```

with `rollup` or `vite`.

cookpete:
Published in `v2.11.1`

cookpete:
> This is good, but would also like to see a `module` field in `package.json` which points to `es6/<entry>`

Note that this was reverted in `v2.11.2` because it broke things: #1574

nemosmithasf:
@cookpete I believe you need to add typescript definitions to the `es6` entry

PR #1553, opened 2022-12-06
Replace Vimeo default preview thumbnail

paladosss:
Replace Vimeo default preview thumbnail size from 295x166 to 640

https://developer.vimeo.com/api/common-formats


paladosss:
The solution is like [here](https://github.com/cookpete/react-player/pull/1057), but the constant is not overwritten

cookpete:
Published in `v2.11.1`

Issue #1574, opened 2023-02-08
fix: breaking update v2.11.1


glpecile:
#### Current Behavior

Project doesn't compile.

#### Expected Behavior

To compile correctly.

#### Steps to Reproduce

1. Compile with version `2.11.1`
2. Boom

#### Other Information

Trace:

```
./node_modules/react-player/es6/ReactPlayer.jsx
--
19:45:31.464 | Module parse failed: Unexpected token (140:8)
19:45:31.464 | You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
19:45:31.464 | \|       const { light, playIcon, previewTabIndex, oEmbedUrl } = this.props
19:45:31.464 | \|       return (
19:45:31.464 | >         <Preview
19:45:31.464 | \|           url={url}
19:45:31.464 | \|           light={light}
19:45:31.464 |  
19:45:31.464 | Import trace for requested module:
19:45:31.464 | ./node_modules/react-player/es6/ReactPlayer.jsx
19:45:31.464 | ./node_modules/react-player/es6/index.js
19:45:31.464 | ./components/chats/FloatingVideoPlayer.tsx
19:45:31.465 | ./components/chats/DisplayFloatingVideoPlayer.tsx
19:45:31.465 |  
19:45:31.465 | ./node_modules/react-player/es6/players/DailyMotion.jsx
19:45:31.465 | Module parse failed: Unexpected token (116:6)
19:45:31.465 | You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
19:45:31.465 | \|     }
19:45:31.465 | \|     return (
19:45:31.465 | >       <div style={style}>
19:45:31.465 | \|         <div ref={this.ref} />
19:45:31.465 | \|       </div>
19:45:31.465 |  
19:45:31.466 | Import trace for requested module:
19:45:31.466 | ./node_modules/react-player/es6/players/DailyMotion.jsx
19:45:31.466 | ./node_modules/react-player/es6/players/index.js
19:45:31.466 | ./node_modules/react-player/es6/index.js
19:45:31.466 | ./components/chats/FloatingVideoPlayer.tsx
19:45:31.466 | ./components/chats/DisplayFloatingVideoPlayer.tsx
19:45:31.466 |  
19:45:31.466 | ./node_modules/react-player/es6/players/Facebook.jsx
[...]

PR #1575, opened 2023-02-10
stream-chat-react loader compatibility

rushi-nonstopio:
image


cookpete:
@rushi-nonstopio I think the issue you are seeing is #1574 and these changes won’t have any effect (the `jsx` files are generated by `npm run build:es6` and are not touched by webpack.

PR #1577, opened 2023-02-10
stream-chat-react loader compatibility

rushi-nonstopio:


rushi-nonstopio:
Already taken care in PR #1575

PR #1588, opened 2023-02-27
Added Vimeo Muteability for iOS

LucidNinja:
Vimeo now has `getMuted` and `setMuted` methods, so we don't need to rely on setting the volume. https://github.com/vimeo/player.js#setmutedmuted-boolean-promiseboolean-error


cookpete:
Nice! Thanks.

PR #1599, opened 2023-03-17
Update the list of supported YouTube domains

Zelenetska:
Added YouTube education support

### DIFF
--- a/src/players/YouTube.js
+++ b/src/players/YouTube.js
@@ -29,7 +29,7 @@ export default class YouTube extends Component {
 
   load (url, isReady) {
     const { playing, muted, playsinline, controls, loop, config, onError } = this.props
-    const { playerVars, embedOptions } = config
+    const { playerVars, embedOptions, host } = config
     const id = this.getID(url)
     if (isReady) {
       if (MATCH_PLAYLIST.test(url) || MATCH_USER_UPLOADS.test(url) || url instanceof Array) {
@@ -42,6 +42,11 @@ export default class YouTube extends Component {
         endSeconds: parseEndTime(url) || playerVars.end
       })
       return
+    }
+    const getHost = () => {
+      if (host) return host
+      if (MATCH_NOCOOKIE.test(url)) return NOCOOKIE_HOST
+      return undefined
     }
     getSDK(SDK_URL, SDK_GLOBAL, SDK_GLOBAL_READY, YT => YT.loaded).then(YT => {
       if (!this.container) return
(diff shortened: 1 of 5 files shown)


CarlosLleraMartin (approved):

src/players/YouTube.js
```diff
@@ -43,6 +43,11 @@ export default class YouTube extends Component {
       })
       return
     }
+    const getHost = () => {
+      if (host) return host
+      if (MATCH_NOCOOKIE.test(url)) return NOCOOKIE_HOST
```
todd-patterson, line 48: regarding lines 47-48 and introducing a new `host` param -

It seems preferable to align on a single paradigm, rather than a mixture of:
1) original: derive the HOST from `url`
2) new: directly set HOST via new param

If we introduce the mixed model, it will be slightly higher cognitive load to maintain and subtle bugs could occur in the future if the basename from input `url` differs from specified `host` param.

Pros of standardizing on # 1 ("derive" approach):
+ Backwards compatible with existing usages of the library and any developers using "youtube-nocookie.com" urls
+ Avoids potential mismatch between `url` basename and the (newly proposed) `host` param

Pros of standardizing on # 2 ("set" approach):
+ Decouples the library from needing to know about specific YouTube domains. New domains could more easily be added in the future without any code-changes, library could just blindly pass-through. (* caveat below)
+ Declarative setting of `host` is debatably clearer than deriving behind-the-scenarios, although I realize many other values are derived from the url (debatably those other values can and should be passed-in as individual, separate params).

* I realize now that the `canPlay` function in patterns.js does perform regex validation against known, preset domains. This has some downsides, since at least with the YT player (and I'd imagine other video players) the list of validate url patterns will continue to evolve and this library will need to be updated accordingly. That's a higher maintenance cost than just allowing the library to pass-through directly to the player, and let the player error-out if the url is malformed or refers to an unsupported path. Removing the `canPlay` paradigm would be a bigger change though, and I realize is out-of-scope. Something to consider for future evolution of this lib.

I think there are merits to both approaches, but seems like #1 should win-out due to:
+ backwards compatible
+ we need to update the regex in patterns.js for each YT domain anyhow
+ avoids potential divergence between `host` and `url` basename

So, what do you think about just mimicking the NOCOOKIE approach and not exposing a new `host` param? (I realize this is contrary to what our teams discussed over email, but at that time  I had only done a very quick read-through of the code and missed the `canPlay` logic.)

todd-patterson (approved):
Approving, with the assumption that my comment will be addressed. Thanks!

PR #1603, opened 2023-03-28
feat: Add Spotify track/episode support

desmaraisp:
Using the [Spotify Iframe API](https://developer.spotify.com/documentation/embeds/references/iframe-api), we can embed and control spotify tracks and episodes. Some things can't be controlled, like playback speed and volume, but otherwise, toggling pause/play, setting the timestamp, settings event callbacks are all things we can do

To use this feature, pass a link of format "spotify:track:id" or "spotify:episode:id" to the player. (Let me know if you want that changed to URLs, I felt like to was more natural to use the spotify URIs instead of URLs, as that's what the spotify api returns most of the time)

I also fixed a small issue related to the loop mechanism being broken when using lazy players.

Small note though. To view items, the viewer MUST be logged in with their spotify account (premium not required, unlike the Spotify playback SDK) otherwise they get a 30-sec preview. Functionality is mostly maintained, the clip is simply cut short

Should partially address #161

I haven't updated the readme yet, wasn't sure if you preferred to do that yourself. Just let me know if you'd like me to do that too

### DIFF
--- a/src/Player.js
+++ b/src/Player.js
@@ -213,7 +213,7 @@ export default class Player extends Component {
 
   handleEnded = () => {
     const { activePlayer, loop, onEnded } = this.props
-    if (activePlayer.loopOnEnded && loop) {
+    if ((activePlayer?._result?.loopOnEnded || activePlayer?.loopOnEnded) && loop) {
       this.seekTo(0)
     }
     if (!loop) {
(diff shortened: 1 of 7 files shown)


dgbogi564:
Not sure if it's applicable for Iframes, but vencord has a [plugin that can control the volume of Spotify embeds](https://vencord.dev/plugins/FixSpotifyEmbeds); it might be worth looking into.

PR #1620, opened 2023-04-26
Add keepPlaying to seekTo

LoyalPotato:
I just added the option to the Youtube player as that is the only one that has a pause on the seekTo.

Made it default to false so that it does not alter any behaviour that is counting on that pause.

Let me know if the other players should have the same option or not. If so, would it be better to keep them as true since it didn't have a behaviour of pause before?


luigibarbato:
It would be great for the other kind of players! 💯

Issue #1598, opened 2023-03-17
Unrecognized request for current URL with undefined at the end


YuriiRysyn:
Be sure to [search for your issue](https://github.com/CookPete/react-player/issues) before opening a new one.
Hi, need some help with. 
#### Current Behavior

For some reason, on a production build, only on a production, even if build prod version local, there is no this issue.  React Player,  before start loading and creating a vidoe blob, from hls streaming, sent request to current URL but add undefined and the end of URL. And this is ofcourse breack logic. I don't understand what this request is. Why it is there, as there should be like that. On different parts of App with different URL, it`s making same. Send some request for current URL with undefined at the end. But only on production.
On local version it is working as expected, there is no this requests at all
For some reason, on a production build, only on a production, even if build prod version local, there is no this issue.  React Player,  before start loading and creating a vidoe blob, from hls streaming, sent request to current URL but add undefined and the end of URL. And this is ofcourse breack logic. I don't understand what this request is. Why it is there, as there should be like that. On different parts of App with different URL, it's making same. Send some request for current URL with undefined at the end. But only on production.
On local version it is working as expected, there is no this requests at all
#### Expected Behavior

Understand what those requests are and fix the issue.
#### Steps to Reproduce

1.
1.
1.

#### Environment
- URL attempting to play: 
- Browser: 
- Operating system: 
- jsFiddle example: https://jsfiddle.net/sv5x3ug1

#### Other Information

![1](https://user-images.githubusercontent.com/67281495/225899174-a497515c-473d-4c0d-a1f3-0ea4351faa3a.png)
![2](https://user-images.githubusercontent.com/67281495/225899179-4325fd23-d188-4e61-9fb0-ab4426578b54.png)

YuriiRysyn:
Updates. This is probably somth with React Player version or maybe npm. Because righ now I have updated React Plyer to latest v, and started getting this errors locally also. 
Little bit strange, as on Prod was same previous version of React Player 
![Screenshot_8](https://user-images.githubusercontent.com/67281495/225903952-8990c585-3f4e-4b34-a536-1d7d0bb9a49f.png)
![Screenshot_7](https://user-images.githubusercontent.com/67281495/225903955-1f4e111a-903e-4a6b-95cd-34e8d85f5bbc.png)

YuriiRysyn:
![Screenshot_2](https://user-images.githubusercontent.com/67281495/225904858-7be41d37-937e-4c96-9934-d8797226a372.png)

[...]

PR #1622, opened 2023-04-26
fix: default player src to empty string to avoid unwanted network requests when using HLS content.

ThatTobMate:
**Issue:**

https://github.com/cookpete/react-player/issues/1598

[This](https://github.com/cookpete/react-player/blob/master/src/players/FilePlayer.js#L27) line is causing unwanted network requests when using a hls source. This line sets the active player `src` to undefined which gets set on the underlying video element as `src="undefined"` this then generates the network requests mentioned above as this defaults to `https://CURRENT_URL/undefined`.

Screenshot 2023-04-26 at 19 07 01

**Fix:**

Default to setting this.player.src as an empty string if the source value is undefined on mount. This is the case when using hls, dash or other media streams. See [here](https://github.com/cookpete/react-player/blob/master/src/players/FilePlayer.js#L320). By defaulting the src value in this case to a empty string the src property gets set on the `video` element as `<video src />`, not ideal but better than `<video src="undefined" />` and it doesn't make any unwanted requests. In the case of HLS streams the HLS package then takes control of this element and loading the actual source anyway.

I tested this by overwriting `this.player.src` to the empty string after it's been set to undefined in the breakpoint above. Seems to work ok and no unwanted network requests being fired.

I was initially going to do this instead:

```
    const source = this.getSource(this.props.url);
    if (source) {
      this.player.src = source;
    }
```

However I noted the comment about strict mode and ensuring the src value is set.

### DIFF
--- a/src/players/FilePlayer.js
+++ b/src/players/FilePlayer.js
@@ -24,7 +24,7 @@ export default class FilePlayer extends Component {
   componentDidMount () {
     this.props.onMount && this.props.onMount(this)
     this.addListeners(this.player)
-    this.player.src = this.getSource(this.props.url) // Ensure src is set in strict mode
+    this.player.src = this.getSource(this.props.url) || '' // Ensure src is set in strict mode
     if (IS_IOS) {
       this.player.load()
     }


duranmla:
Side comment this happens to me when I add multiple sources. I was trying to get the video element behavior of adding several sources to fallback until the browser can play and I start to have this `src="undefined"` check the example here:

https://codesandbox.io/embed/loving-mcclintock-vefv65?fontsize=14&hidenavigation=1&theme=light

cookpete:
I’m hoping this is fixed by #1648. If not please let me know.

Issue #1631, opened 2023-05-07
Deno support (for e.g. Fresh)


arnemolland:
#### Current Behavior

Using any v2.x.x `ReactPlayer` component results in errors when rendered with Fresh:
```bash
[Error] Error: Missing Suspense. The throwing component was: ReactPlayer
	(anonymous function) (main.js:8:842)
	I (chunk-HGUCLWZ2.js:1:7033)
	ce (chunk-HGUCLWZ2.js:1:8896)
	walk (main.js:43:2032)
	walk (main.js:43:2233)
	revive (main.js:43:2242)
	Module Code (520775:39:1800)
```

This is largely due to react-player forcing the use of Suspense as long as it's rendered in the browser. This approach works fine for _most_ frameworks, but since Fresh's render method is synchronous and _non-streaming_ it doesn't support Suspense boundaries.

#### Expected Behavior

I would suggest adding an option to disable Suspense, for cases where Suspense is not supported at all. Currently Suspense is always enabled;

https://github.com/cookpete/react-player/blob/662082a2f2edc863e8ca9cefdd7a3c88ad49ea0d/src/ReactPlayer.js#L174-L180

as long as the component is rendered in the browser:

https://github.com/cookpete/react-player/blob/662082a2f2edc863e8ca9cefdd7a3c88ad49ea0d/src/ReactPlayer.js#L12-L18

#### Steps to Reproduce

1. `deno run -A -r https://fresh.deno.dev repro-example`
2. Add react-player to import-map, e.g. 
```json
{
  "imports": {
    "react-player": "https://esm.sh/react-player@2.12.0?alias=react:preact/compat,react-dom:preact/compat:preact/compat&deps=react,react-dom",
  }
}

```
3. Render any `<ReactPlayer/>`. 

#### Environment
- URL attempting to play:  Any
- Browser: Any
- Operating system: Any
- jsFiddle example: None

luwes:
is this still an issue?

is Deno polyfilling `window`?

see this code in ReactPlayer
https://github.com/cookpete/react-player/blob/64c2130218ae610a1b2ed63647a35530fa7c0a97/src/ReactPlayer.js#L12-L18

arnemolland:
> is this still an issue?
> 
> is Deno polyfilling `window`?
> 
> see this code in ReactPlayer
> 
[...]

PR #1632, opened 2023-05-07
Deno compatibility

arnemolland:
Draft/suggestion for #1631. The `typeof document` check is needed for Deno environments, as `window` is part of the runtime API.


luwes (approved):
LGTM! thank you for the contribution

PR #1639, opened 2023-05-27
Adding keepPlaying to other player types

LoyalPotato:
This is a continuation on the other pull request ( #1620 ) that added this only to YouTube players.

Also changed the default values from the generic player into each specific player, as YouTube for example, has a different default than the rest.


heff:
@LoyalPotato, new maintainer here and getting ramped up. Can you help me understand, is this changing any of the existing behavior or keeping it? Does seekTo have different results depending on which player is used?

LoyalPotato:
@heff I believe it's more of an addition of behavior. I kept the default values for the new parameter the same as if they weren't there (so most of them are true since it didn't have any option to pause before).
The YouTube player already had this functionality, but it didn't have an option to allow the video to keep playing after using `seekTo` (#1620). What I've done here is just extend the functionality of that PR into the other players.

heff:
Great, makes sense. Thanks!

PR #1648, opened 2023-06-23
Media can not be played if `url` is an array for the file player

Losses:
If the `url` is an array, and the player is a file player, the `src` property will become `"undefined"`, this will trigger an error and cause the media unable to be played.

image


mezohaib (approved):

mezohaib:
Why this has not been merged

PR #1654, opened 2023-07-25
CI

jessp01:
Use GH actions for CI.

All tests pass with node 16 but the `build` jobs seem to fail with higher versions. Please see https://github.com/jessp01/react-player/actions/runs/5662727275/job/15343189737 for an example of the failure.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,7 +21,7 @@ jobs:
     - run: yarn lint
     - run: yarn test:coverage -v
     - run: yarn build:lib
-    - run: yarn build:demo
+    # - run: yarn build:demo
     - run: yarn build:dist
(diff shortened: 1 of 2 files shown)


heff:
LGTM, thanks @jessp01! I ran into the same node version issue when setting up the repo for the first time and [found this answer](https://stackoverflow.com/questions/73144960/error-error0308010cdigital-envelope-routinesunsupported-at-new-hash-nodei). But we should figure out how to get the build up to date.

jessp01:
Cheers, @heff. Now that you've merged it, I can have a look at the build issue. While we're conversing, any thoughts on https://github.com/cookpete/react-player/pull/1653?

jessp01:
@heff , please see https://github.com/cookpete/react-player/pull/1664

PR #1664, opened 2023-08-30
Override default webpack hashing algo

jessp01:
This is submitted as a follow-up to this pull: https://github.com/cookpete/react-player/pull/1654

See https://nodejs.org/en/blog/release/v17.0.0#openssl-30

Since we don't want to require that the `node` binary be built against OpenSSL 3 (as it narrows down our supported ENVs matrix substantially) we had several options:

- set `NODE_OPTIONS="--openssl-legacy-provider"`
- override default webpack hashing algo

I went with the latter because:
a. MD4 is rubbish anyways
b. NodeJS 16 does not recognise the` --openssl-legacy-provider` option and so, if set, the build fails with it


heff:
Awesome, thanks @jessp01. Is it necessary to check in the dist/ builds with this change? It's causing a merge conflict and the contributing guide says you generally shouldn't.

jessp01:
Hi @heff ,

Since looking through the diffs of these minimised files will no doubt prove to be painful and seeing how the build and tests all passed, I reckon it'd be safe to simply override the existing files but I leave the decision to you, of course.

cookpete:
Updating the `dist` files is part of the `npm version` and `npm publish` flow. There’s no need to commit them any other time.

https://github.com/cookpete/react-player/blob/master/CONTRIBUTING.md#dist-files

jessp01:
Hi @cookpete ,

Shall we add these to `.gitignore` to avoid this in future? For now, they can just be reverted to their last version.

luwes:
thanks for the contribution! 

with this larger refactor it's not an issue anymore. closed by https://github.com/cookpete/react-player/pull/1684

jessp01:
Hi @luwes ,

This bit is still relevant: https://github.com/cookpete/react-player/pull/1664/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR12

Cheers,

PR #1684, opened 2023-10-03
fix: modernize build using esbuild

luwes:
Webpack is not a great fit for bundling libraries and the version 4 that is used here is pretty outdated, this PR changes the build commands to use esbuild which is a lot faster and does not require Babel nowadays.

Webpack is better suited for the demo app but it's so minimal that we can get away with using esbuild (serve) and plain CSS instead of SASS. This will allow us to remove a bunch of dependencies and configuration.

- [x] esbuild library bundles 
- [x] esbuild demo with new build process
- [x] remove Babel w/ possible test tool replacement
- [x] verify bundles are working as intended
  verified locally by pasting the iife bundles in the readme js fiddles 
  we might have to do a canary/beta release first to verify bundling


cookpete (approved):
This looks awesome. Nice work!

PR #1693, opened 2023-10-16
Fix #1692: Error event not triggering when using array of urls

### LINKED ISSUE #1692: Error event not being trigger when using url as array
#### Current Behavior
The video player stays on error state without triggering error event

#### Expected Behavior
I should get the error event called so I can handle it as expected. On the code sandbox below, I should receive an error on the video player with multiple urls as I do with the one with only one url. 

#### Steps to Reproduce
1. Try to attach an error callback to a video with multiple wrong urls
2. Your callback never gets fired

#### Environment
https://codesandbox.io/s/condescending-water-m6qdjr?file=/src/App.js

anampartho:
Aims to fix issue: #1692 
For array of URLs, we are rendering `<source />` tags. Errors on these tags do not trigger the parent players error event. So created a `handleSourceError` function which simply creates an `Error event` and dispatches that event on the player.

### DIFF
--- a/src/players/FilePlayer.js
+++ b/src/players/FilePlayer.js
@@ -330,11 +330,19 @@ export default class FilePlayer extends Component {
     return url
   }
 
+  // As errors on the `source` tag does not trigger the error event of the
+  // parent video/audio tag (player). We have to manully trigger
+  // the players error event.
+  handleSourceError = () => {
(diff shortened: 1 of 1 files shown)


luwes:
This is outdated and in v3 you will be able to listen directly to the source error events. Thanks for the contribution.

PR #1705, opened 2023-10-27
Add aria-label to Preview component

schdaniel20:
In this commit, an `aria-label` attribute has been added to the Preview component. This change ensures that the Preview component is now more compliant with accessibility standards.


luwes (approved):
thank you, LGTM!

PR #1735, opened 2024-01-31
docs: fix badges

okay-head:
Remove unwanted underline between shield badges.

Possibly a result of how line endings are handled in github flavored markdown

The issue:
image

The fix:
image


luwes:
Nice! Thank you!

PR #1741, opened 2024-02-20
fix/display prop handling in render method

Ritika1828:


luwes:
thanks but this is not a bug. 
it's a shorthand for `display: display`

Ritika1828:
if display is an object and if we are passing it as a prop it will be appending like object
example - 
const display={
  fontSize:"14px",
  bgColor:"red",
}

const style = {
      width: '100%',
      height: '100%',
      display
    }
    

output will be 
{
  width: '100%',
  height: '100%',
  display: { name: 'Ritika', bgColor: 'red' }
}
which will not correctly applied to the styles that we are passing to div
the correct should be by using spread operator in display while using it in style

luwes:
this is display I believe https://developer.mozilla.org/en-US/docs/Web/CSS/display

Ritika1828:
can we pass more styles by passing any style props ?

PR #1805, opened 2024-04-19
Fix where onReady not firing because IOS doesn't load past metadata with autoplay off

kozr:
### Bug Fix

There is a problem where the `onReady` function in `react-player` isn't being called on IOS when `autoPlay` is `false`. This is because IOS stops loading past metadata when autoplay is off (`readyState` remains at 1).

Also get rid of line 83 which seems like a duplicate?

[Stackoverflow](https://stackoverflow.com/questions/9920297/html5-audio-canplay-event-doesnt-fire-on-safari-mac-desktop)
[GitHub Livepeer PR](https://github.com/livepeer/ui-kit/pull/342)


kozr:
@luwes

luwes:
thanks for the contribution but this needs some more investigation. If preload=none even loadedmetadata would not fire. 
I'll have to dig in when onReady is called in all players.

luwes:
Closing as this is outdated with v3.

Issue #1613, opened 2023-04-16
Types wrong for ESM + module: Node16 / NodeNext 


karlhorky:
Be sure to [search for your issue](https://github.com/CookPete/react-player/issues) before opening a new one.

#### Current Behavior

The types for `react-player@2.12.0` are incorrect when using ESM (eg. `"type": "module"` in `package.json`) and modern Node.js module resolution (`"module": "Node16"` or `"module": "NodeNext"` in `tsconfig.json`)

See the error on [Are The Types Wrong?](https://arethetypeswrong.github.io/?p=react-player%402.12.0)

```
The CJS module resolved at the package under contains a simulated export default with an __esModule marker, but no top-level module.exports. Node does not respect the __esModule marker, so accessing the intended default export will require a .default property access in Node from an ES module.
```

![Screenshot 2023-04-16 at 18 17 56](https://user-images.githubusercontent.com/1935696/232326143-2602b610-8869-41ff-8b29-1fb26ca0a9a1.png)

You can also see this on [this TypeScript Playground](https://www.typescriptlang.org/play?target=99&moduleResolution=99&module=100#code/JYWwDg9gTgLgBAJQKYEMDGMAKAbFBPJKOAMyghDgHIpUMBaMXAqSgKFYHoPEkA7AE0JwUcAJoQArgBUJAIyRwAbsEEQ4jfIVYAeZOixMhEqNgC8lABYwYYAM4AuLgHcXAOjySYcpK7TkOTigwaBYA-IqmeLYAXgCsAMoAbJgAqgCydCmUcBwAfEA), with the code from the first example in the readme. Error message:

```
JSX element type 'ReactPlayer' does not have any construct or call signatures.(2604)
```

#### Expected Behavior

There should be no error

#### Steps to Reproduce

1. Visit [this TypeScript Playground](https://www.typescriptlang.org/play?target=99&moduleResolution=99&module=100#code/JYWwDg9gTgLgBAJQKYEMDGMAKAbFBPJKOAMyghDgHIpUMBaMXAqSgKFYHoPEkA7AE0JwUcAJoQArgBUJAIyRwAbsEEQ4jfIVYAeZOixMhEqNgC8lABYwYYAM4AuLgHcXAOjySYcpK7TkOTigwaBYA-IqmeLYAXgCsAMoAbJgAqgCydCmUcBwAfEA)
2. See the error

#### Environment
- URL attempting to play: --
- Browser: --
- Operating system: --
- jsFiddle example: --

[...]

PR #1828, opened 2024-04-21
Move exports.key.default to last position

karlhorky:
Addresses https://github.com/cookpete/react-player/issues/1613#issuecomment-2067993523

---

Comment content copied here:

Ah, seems like this `'react-player/file'` import fails with a `Default condition should be last one` error message on compile time (Next.js) though 😬

```
Module not found: Default condition should be last one
  5 | // eslint-disable-next-line import/no-unresolved -- eslint-plugin-import doesn't understand `exports` in package.json yet https://github.com/import-js/eslint-plugin-import/issues/1810
> 6 | import ReactPlayer from 'react-player/file';
    | ^
  7 |
  8 | // ```
  9 |

https://nextjs.org/docs/messages/module-not-found
```

So maybe [the `"exports"` config in `package.json`](https://unpkg.com/browse/react-player@3.0.0-canary.3/package.json) is misconfigured...

Eg. an example PR fixing such an error:

- https://github.com/firebase/firebase-js-sdk/pull/7007

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,16 +19,16 @@
   "type": "module",
   "exports": {
     ".": {
-      "default": "./dist/index.js",
-      "types": "./dist/index.d.ts"
+      "types": "./dist/index.d.ts",
+      "default": "./dist/index.js"
     },
     "./lazy": {
-      "default": "./dist/lazy/index.js",
-      "types": "./dist/lazy/index.d.ts"
+      "types": "./dist/lazy/index.d.ts",
+      "default": "./dist/lazy/index.js"
     },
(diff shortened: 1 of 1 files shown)


luwes (approved):
thank you, LGTM! lets try this

karlhorky:
Glad to help, thanks for the review and merge!

I see that this was published in [`react-player@3.0.0-canary.4`](https://www.npmjs.com/package/react-player/v/3.0.0-canary.4), thanks!

I can confirm that this is working for our Next.js project 👍

PR #1833, opened 2024-04-27
add an option to load libraries from custom domain

maxired:
Hi, this Pr is an attempts to provide new configuration options to allow to load hls, dash and flv library from custom domain.
I am required to do so to use that library in a controller environment where specific domains need to be allowed to fetch content.


vkrbt:
Have the same issue and need I this changes also

luwes:
This is outdated in v3. The players are not loaded via CDN anymore but via local NPM imports with splitting.

PR #1841, opened 2024-05-08
Fix #1460: Remove memorization of getConfig to prevent stale config renders

### LINKED ISSUE #1460: File config poster attribute not changing with new state
Hi Dear,
I am using React Player as file player with attribute of poster which is set to State. 
The poster is not changing when the State of poster is changed where "imageToView" is the url of poster image

 config={{
              file: {
                attributes: {
                  poster: ImageToView
                    
                },
              },
            }}

Thanks for your support

ReaZzy:
This PR addresses #1460 by removing the memorization of `getConfig` in `src/ReactPlayer.js`. This change prevents stale config renders by ensuring that the latest configuration is always used.

Changes:
- Updates `getConfig` to be a non-memoized function.

### DIFF
--- a/src/ReactPlayer.js
+++ b/src/ReactPlayer.js
@@ -119,7 +119,7 @@ export const createReactPlayer = (players, fallback) => {
       return null
     })
 
-    getConfig = memoize((url, key) => {
+    getConfig = (url, key) => {
       const { config } = this.props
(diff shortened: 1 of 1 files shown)


luwes:
memoization is removed in v3. Thanks for the contribution.

PR #1853, opened 2024-06-06
Update default hlsVersion and dashVersion in README.md

federaljules:
README had the old default version numbers for `hlsVersion` and `dashVersion`, so this PR is to update those. Copied the new version numbers from `react-player/src/props.js`


luwes:
Please use https://github.com/cookpete/react-player#sdk-overrides

PR #1859, opened 2024-06-26
feat: BlueBillywig player

jkoppen-bluebillywig:
Ads support for Blue Billywig media (live or on-demand, interactive, playlists) and monetization.


luwes:
thanks for the contribution but this player won't be merged into the project as the user base is too little.
it's will work great as a custom player https://github.com/cookpete/react-player#adding-custom-players though.

we'll soon create a list on the readme with 3rd party custom players for discoverability!

PR #1875, opened 2024-08-27
Support for Gumlet's player

mudit2108:
This update introduces support for embedding Gumlet player using an iframe tag.

Thank you for your attention and consideration.

### DIFF
--- a/examples/react/src/App.js
+++ b/examples/react/src/App.js
@@ -357,6 +357,12 @@ class App extends Component {
                 </td>
               </tr>
               <tr>
+                <th>Gumlet</th>
+                <td>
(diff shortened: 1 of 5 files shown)


luwes:
@mudit2108 thanks for your contribution! 

This will have to move to a media element if possible. 
I saw the API is based on https://github.com/embedly/player.js 

Maybe we could get it to work with a more generic player.js video element?
https://github.com/muxinc/media-elements/issues/132

adityapatadia:
Hey @luwes  what can we do here? Should we send PR again?
I see that issue of `media-elements` is still open.

adityapatadia:
@luwes @cookpete can you let us know if you are willing to accept PR for this? 

I also see there is no activity on this project. If needed, we at @gumlet can help to make this package better and lend a helping hand. We not only want to add Gumlet support but also want to see this player have all provider support it had in 2.x

PR #1891, opened 2024-10-09
Update default versions on readme

appare45:
Updated the default versions of hls.js and dash.js on readme based on props.js

https://github.com/cookpete/react-player/blob/3cda23fd0d43e747b8b8d2b58f04ceb6173b3552/src/props.js#L185-L186

Props are used here

https://github.com/cookpete/react-player/blob/3cda23fd0d43e747b8b8d2b58f04ceb6173b3552/src/players/FilePlayer.js#L163-L201


luwes:
it's now recommended to do this with version overrides in v3 
see here https://github.com/cookpete/react-player#sdk-overrides

Issue #1428, opened 2022-03-30
React 18 NextJS 12.1 - Hydration failed


davidkhierl:
#### Current Behavior
causing error: Hydration failed because the initial UI does not match what was rendered on the server.

#### Expected Behavior
should not cause error

#### Steps to Reproduce
1. Fresh install nextjs app, will install react 18 by default
1. include ReactPlayer component

JulianLeviston:
We got around this by effectively memoizing the loading state of it.

sandrooco:
I‘d recommend to check on something like window (which is only available on client side). This needs to be done in useEffect for proper handling.
Then you simply conditionally display the player.

`hasWindow && <ReactPlayer …/>`

Dovalization:
Like @sandrooco said, this can fix the issue.
Just a small component to show the full picture 

![video_player_component](https://user-images.githubusercontent.com/66652218/162261217-b6a1c410-5457-4772-b45b-c545a0104741.png)

inderrr:
Dynamically import the react-player to remove this error.
This fixed the error for me.

```javascript

import dynamic from "next/dynamic";
const ReactPlayer = dynamic(() => import("react-player/lazy"), { ssr: false });

```

cookpete:
These workarounds look good. It’s also possible this is fixed in `react@18.1.0` and `react-dom@18.1.0`

degitgitagitya:
Um sorry, why this issue closed? I'm still getting hydration error with `react@18.1.0` , `react-dom@18.1.0` , `next@12.6.0`.

Dynamic import fixed the issue, but the `onProgress` props isn't working.

inderrr:
@degitgitagitya It seems to be a issue with `react-player@2.10.1`. I downgraded to `2.10.0` and `onProgress` worked again.
I tested with `react@18.1.0` , `react-dom@18.1.0` , `next@12.1.6`

mrgnou:
Rendering the component on the client side only isn't a solution and has negative performance impact, this issue should be reopened.

MuMaestro:
Is it an alternative is to include a empty iframe ? I suppose iframes are a good solution as they could be marked optionaly as unoptimized. Or maybe it could include a option for client-only, where it waits of an effect to set a boolean true if window is defined, it does defeat the pourpouse of optimization, but makes a transparent workaround for some providers.
(I'm saying thins whithout a look at source code)

mrgnou:
[...]

PR #1912, opened 2025-01-10
fix issue where setState is called before hydration completes

gmferise:
Fixes the following exception:
```
This Suspense boundary received an update before it finished hydrating. This caused the boundary to switch to client rendering. The usual way to fix this is to wrap the original update in startTransition.
```


gmferise:
Related: #1428

gmferise:
Requesting review from a maintainer.

@luwes seems to be most active?

PR #1933, opened 2025-05-18
Fix: prevent infinite recursion in deepmerge by extracting children prop

xkelxmc:
- Updated ReactPlayer to destructure and exclude children from props before merging with defaultProps.
- This prevents deepmerge from causing infinite recursion when children are present.
- Children are now explicitly passed to the underlying player component, enabling custom content without recursion issues.

### DIFF
--- a/src/ReactPlayer.tsx
+++ b/src/ReactPlayer.tsx
@@ -33,7 +33,7 @@ export const createReactPlayer = (players: PlayerEntry[], playerFallback: Player
     return null;
   };
 
-  const ReactPlayer: ReactPlayer = React.forwardRef((_props, ref) => {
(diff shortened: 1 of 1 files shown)


luwes (approved):
LGTM, thanks!

src/ReactPlayer.tsx
```diff
@@ -87,6 +87,7 @@ export const createReactPlayer = (players: PlayerEntry[], playerFallback: Player
             ? { display: 'block', width: '100%', height: '100%' }
             : { display: 'block', width, height, ...style }}
           config={config}
+          children={children}
         />
```
luwes, line 91:
```suggestion
        >{children}</Player>
```

PR #1959, opened 2025-07-15
fix: muted prop bug

### LINKED ISSUE #1957: Muted state doesn't work on HLS Player for v3
Issue: We can pass the `muted` state to the React Player component and pass m3u8 URL, which renders the `<hls-player>` element, but the `muted` state doesn't work.

I noticed this issue when updating to v3 on my codebase, and it also happens in the demo link: https://cookpete.github.io/react-player/  

I click the muted checkmark, but the video is not muted. 

The muted state is properly set on the on click handler and passed as props to the video player, but the prop has no effect.

Image

Image

luwes:
fix #1957

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,16 +9,16 @@
       "version": "3.2.0",
       "license": "MIT",
       "dependencies": {
-        "@mux/mux-player-react": "^3.5.0",
-        "cloudflare-video-element": "^1.3.2",
-        "dash-video-element": "^0.1.5",
+        "@mux/mux-player-react": "^3.5.1",
+        "cloudflare-video-element": "^1.3.3",
+        "dash-video-element": "^0.1.6",
         "deepmerge": "^4.0.0",
-        "hls-video-element": "^1.5.5",
-        "spotify-audio-element": "^1.0.1",
-        "twitch-video-element": "^0.1.0",
-        "vimeo-video-element": "^1.5.1",
-        "wistia-video-element": "^1.3.2",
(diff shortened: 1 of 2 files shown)


cduff:
@luwes I just tried upgrading `react-player` to 3.2.1 and when it tries to play a Vimeo video the following exception is thrown:

```
player.es.js:2290  Uncaught (in promise) TypeError: There must be a value to set.
    at eval (player.es.js:2290:1)
    at new Promise (<anonymous>)
    at Player.set (player.es.js:2287:1)
    at Player.setMuted (player.es.js:3111:1)
    at eval (vimeo-video-element.js:373:1)
```

I'm guessing it's caused by this change?

luwes:
thanks for reporting, that looks like a bug in https://github.com/vimeo/player.js
could you provide a codesandbox?

cduff:
> thanks for reporting, that looks like a bug in https://github.com/vimeo/player.js could you provide a codesandbox?

Sure @luwes, here you go, minimal reproduction:
https://codesandbox.io/p/sandbox/react-player-vimeo-muted-bug-5xr58p

Basically just:

```ts
export default function App() {
  return <ReactPlayer src="https://vimeo.com/90509568" />;
}
```

cduff:
> thank you! fixed in v3.3 https://codesandbox.io/p/sandbox/react-player-vimeo-muted-bug-forked-kh34q9?file=%2Fsrc%2FApp.tsx

Thanks @luwes!

PR #1970, opened 2025-07-22
fix: prevent React warnings for unknown event handler properties

ryanburst:
Addresses React warnings about "Unknown event handler property" for ReactPlayer-specific event handlers (onReady, onStart, onPlay, onPause, onEnded, onLoadStart) by filtering them out before passing props to the underlying HTML video element.

The issue was that all props starting with "on" were being passed down to the underlying player component, including ReactPlayer-specific event handlers that don't exist on HTML video elements.

This fix filters out ReactPlayer-specific event handlers while preserving standard HTML video event handlers, eliminating console warnings without breaking functionality.

Added test to verify the fix works correctly.

### DIFF
--- a/src/Player.tsx
+++ b/src/Player.tsx
@@ -71,7 +71,7 @@ const Player: Player = React.forwardRef((props, ref) => {
   // to the underlying HTML video element, which causes React warnings about unknown
   // event handler properties
   const eventProps: Record<string, EventListenerOrEventListenerObject> = {};
-  const reactPlayerEventHandlers = ['onReady', 'onStart', 'onPlay', 'onPause', 'onEnded', 'onLoadStart'];
+  const reactPlayerEventHandlers = ['onReady', 'onStart'];
 
   for (const key in props) {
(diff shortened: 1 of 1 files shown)


src/Player.tsx
```diff
@@ -67,10 +67,14 @@ const Player: Player = React.forwardRef((props, ref) => {
     return null;
   }
 
+  // Filter out ReactPlayer-specific event handlers to prevent them from being passed down
+  // to the underlying HTML video element, which causes React warnings about unknown
+  // event handler properties
   const eventProps: Record<string, EventListenerOrEventListenerObject> = {};
+  const reactPlayerEventHandlers = ['onReady', 'onStart', 'onPlay', 'onPause', 'onEnded', 'onLoadStart'];
```
luwes, line 74:
```suggestion
  const reactPlayerEventHandlers = ['onReady', 'onStart'];
```

`'onPlay', 'onPause', 'onEnded', 'onLoadStart'` are valid HTMLMediaElement events

luwes (approved):
thanks 👍

PR #1973, opened 2025-07-31
Fix: Add share URLs to TikTok pattern matching

AshesOfOwls:
This is a PR which should depend upon https://github.com/muxinc/media-elements/pull/150 merging.

If there are any updates to the regex in that PR then this should also be updated.

---

TikTok supports two versions of URL's:

https://www.tiktok.com/@_luwes/video/7527476667770522893
and
https://www.tiktok.com/share/video/7527476667770522893

The second format redirects to the first format, but is still accessible as an embed src. This is useful since TikTok users may change usernames.


AshesOfOwls:
@Ritik-gh Perfect, thanks! I will go back and see what else was wrong to make it fail :)

Ritik-gh:
Cool

Ritik-gh:
Awesome!

luwes (approved):
👍

PR #2018, opened 2025-11-23
docs: fix path error in package README.

rxliuli:
I came across this module on npm and am currently evaluating it for use. I noticed a broken link in the README and fixed it while I was at it.
image


dylan-seneca:
Thanks for updating this: I've also been evaluating this package for use in a project and ran into the same broken link.

PR #2041, opened 2026-06-14
fix: prevent Wistia crash when src changes while playing

### LINKED ISSUE #2000: Wistia player crashes when src changes while playing
After updating react-player from v2.16.1 → v3.3.3, I started experiencing a crash when switching the video source while the player is already playing. This issue is also reproducible in the demo from the current repository.

**Steps to reproduce:**

1. Open the demo.
2. Click the "Play" button before opening any video.
3. Open any Wistia video.

**Actual behavior:**
The app crashes with the following error:
`TypeError: Cannot read properties of undefined (reading 'play')`

**Expected behavior:**
The player should switch to the new video source and autoplay it, without throwing an error.

I would be grateful for any guidance on how to handle this or a potential fix from your side.

https://github.com/user-attachments/assets/0b0ae9ed-a86e-4020-b558-4356ef3fb29a

Malaybhai11:
## Problem

After upgrading from v2.16.1 to v3.3.3, switching the video source while playing causes a Wistia player crash (fixes #2000).

## Root Cause

In Player.tsx, the underlying player component (including wistia-video-element) is not re-mounted when src changes. The Wistia custom element keeps internal state (the old player instance) that becomes stale/inconsistent when a new source is set while the previous video is still playing, leading to a crash.

## Fix

Added key={props.src} to the Player component in Player.tsx. This forces React to fully unmount and remount the player element whenever the src prop changes, ensuring the old Wistia player instance is properly destroyed before a new one is created.

This is a one-line change that applies to all player types, providing consistent and safe source-change behavior across Wistia, YouTube, Vimeo, and other providers.

## Changes

- src/Player.tsx: Added key={props.src} prop to the Player element


Malaybhai11:
Hi folks! Any chance someone could take a look at this PR? It fixes a crash in the Wistia player when the src changes while playing — the player was trying to call methods on an already-unmounted instance. The fix adds a simple guard check before accessing the player reference. Would appreciate a review!'

